PR #46, opened 2021-10-28
Migrate from tailwind > doses description list

andycarrell:
Also renames the file because, yeh it was wrong

image

@finnclark lmk if I'm doing it not quite right hehe


andycarrell:
Superseded by #55

PR #54, opened 2021-10-29
Move app into subfolder to prep for mono repo

theramis:
I couldn't get husky install to work on my windows pc. 

So will need someone to add that please.


.github/workflows/lint.yml
```diff
@@ -20,6 +20,6 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
           cache: "npm"
       - name: Install dependencies
-        run: npm install
+        run: npm ci
```
andycarrell, line 23:
```suggestion
        run: npm install
```

I _think_ we wanna keep this as install - `ci` is important for the build process, but not really for eslint, where we're happy for the "other stuff" to happen when we install

If we want to verify that `npm ci && npm run build` work as expected I think it's valuable to add that as an additional workflow. For example, at work I setup:

```
- name: Install dependencies - early feedback on production install
  run: npm ci --production
- name: Build
  run: npm run build
```

app/pages/_document.tsx
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @next/next/no-document-import-in-page
```
andycarrell, line 1:
```suggestion
```

Believe no longer necessary

andycarrell:
@theramis for husky:

https://typicode.github.io/husky/#/?id=does-it-work-on-windows

jishaal:
Move to WSL bro 😏

andycarrell (approved):

PR #55, opened 2021-10-29
Migrate from tailwind > doses description list 

andycarrell:
_Supersedes #46_

_Opening this as draft against #54, should automatically rebase against `main`_


app/components/DosesDescriptionList/DosesDescriptionList.tsx
```diff
@@ -18,23 +22,21 @@ export const DosesDescription = ({
...
-          <CheckIcon className="h-4 w-4" />
+        <div className="check">
+          <CheckIcon />
         </div>
       )}
       <div
-        className={cx("data-text", {
-          ["data-text-complete"]: hasMetTarget,
-        })}
+        className={cx("data-text", { ["data-text-complete"]: hasMetTarget })}
```
andycarrell, line 33: Discussed briefly with Finn - these classes may move or change, but it's unclear atm whether they belong to this component or not. Leaving them out for now!

app/components/DosesDescriptionList/DosesDescriptionList.tsx
```diff
@@ -1,7 +1,11 @@
 import cx from "classnames";
 import { CheckIcon } from "@heroicons/react/solid";
-import { DhbPopulationDoseData } from "../types/VaccineDataTypes";
-import numberFormatter from "../utilities/numberFormatter";
+
+import { DhbPopulationDoseData } from "../../types/VaccineDataTypes";
+
+import numberFormatter from "../../utilities/numberFormatter";
+
+import "./doses-description-list.scss";
```
andycarrell, line 8: Is this how we want to name scss files?

jishaal: I went with `ComponentName.module.scss`, no strong preference

app/components/DosesDescriptionList/DosesDescriptionList.module.scss
```diff
@@ -0,0 +1,41 @@
...
+
+.check {
+  width: 1rem;
+  height: 1rem;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: var(--progress-comlete);
+  background: var(--complete-background);
+  border-radius: 99999px;
```
andycarrell, line 34: What to do about this? - We could reference `./styles.scss`, which we can do by importing the stylesheet OR should the border radius be a css `var`

I don't know the answer hehe

finnclark: There's a css var for this radius value right?

app/pages/index.tsx
```diff
@@ -1,11 +1,12 @@
 import { DateTime } from "luxon";
 import { GetServerSideProps } from "next";
 import { useEffect, useState } from "react";
-import { Progress } from "../components/Progress";
+import { Progress } from "../components/Progress/Progress";
```
andycarrell, line 4: Silly goose!
```suggestion
import { Progress } from "../components/Progress";
```

app/components/DosesDescriptionList/DosesDescriptionList.module.scss
andycarrell: @finnclark not really, it's an `scss` var, but it hasn't been defined on the `:root` so it's not a css variable - do you think these should be?

https://github.com/theramis/aucvid/pull/55/files#diff-6181a4434ec7490436656df1775ef3f851c697d0408c56ace521630c2c7f278eR8

finnclark: oh sorry. I reckon just importing the scss then. The css vars are really only being used for colors atm

andycarrell: Cool, happy with that approach. I'll tidy this up in the next PR, since we can't import the css exactly as is, will need some minor refactoring - nothing too arduous tho 💃

app/components/DosesDescriptionList/DosesDescriptionList.tsx
```diff
@@ -63,7 +67,7 @@ const DosesDescriptionList = ({ dhbData }: DosesDescriptionListProps) => (
       hasMetTarget={dhbData.hasMetSecondDoseTarget}
```
jishaal, line 67: While we are here did we want to go back to `dose` vs `doses`?  Can be seperate 

otherwise LGTM

jishaal (approved):

finnclark (approved):

PR #56, opened 2021-10-29
Removes buggy datetime until we have a more robust solution

andycarrell:


theramis (approved):

app/pages/index.tsx
```diff
@@ -104,10 +101,6 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
                 Ministry of Health NZ
               </a>
             </p>
-            <div className="hidden sm:block">&#8226;</div>
-            <p title={formatTimeTitle(dataFetchedAtTimeUtc)}>
```
jishaal, line 108: Is the bug with `dataFetchedAtTimeUtc` being incorrect intermittently?

andycarrell: The bug is "updated less than an hour ago" never changes, for example if you leave the browser tab open for longer than an hour

jishaal: Oh right! A solution would be to have a  component that takes in the date, run in a useEffect with setInterval to update every hour to solve the staleness problem

jishaal (approved):

app/pages/index.tsx
andycarrell: Yeh that sounds great actually 🕺

PR #57, opened 2021-10-29
Add button to toggle between dark/light mode

jishaal:
@finnclark this is what I was thinking for the dark mode toggle btn. Feedback would be primo. 

Lets hope the preview builds 😅

### DIFF
--- a/app/components/DarkModeToggle/DarkModeToggle.module.scss
+++ b/app/components/DarkModeToggle/DarkModeToggle.module.scss
@@ -1,0 +1,10 @@
+.toggleButton {
+  width: 1.25rem;
+  height: 1.25rem;
+  color: var(--dark-mode-toggle);
+  transition: color 0.2s ease-in-out;
+
+  &:hover {
+    color: var(--dark-mode-toggle-hover);
+  }
+}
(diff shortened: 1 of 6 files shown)


finnclark:
Hey mate, appreciate the effort but if it's ok, I would be keen to think about this more... Icon toggle buttons aren't a great UX pattern as there's confusion between whether the icon represents the state it's in vs what typically a button does which is describe the action it takes

jishaal:
Sure thing @finnclark 👍🏼

app/components/DarkModeToggle/DarkModeToggle.module.scss
```diff
@@ -0,0 +1,10 @@
+.toggleButton {
+  width: 1.25rem;
+  height: 1.25rem;
```
finnclark, line 3:
```suggestion
  width: 2.5rem;
  height: 2.5rem;
  padding: .625rem;
```

finnclark (changes requested):
Suggestion to increase the touch target to 40px.

another ask - could we see what it's like with the icons swapped so the icon represents the action rather than the state?

finnclark (approved):
LGTM for now! Might look at something a little different with new design... I'm personally more a fan of a control that's a bit clearer. But this is a great thing to provide value incrementally. Perfection is the enemy of done or something?

andycarrell (approved):

jishaal:
Thanks @finnclark , fair enough definitely understand this isn't ideal => we can move to the toggle switch or something more clearer once you've had a chance to explore

PR #58, opened 2021-10-30
Add new page and region filter dropdown

jishaal:
- Rough dropdown we can use to filter the data. Will style this further + see if we want to go custom route though rather focus on getting it working
- New page at `/angelfish` and adds the `no-follow` meta tag. We can start building the new functionality on this


app/components/RegionDropdown/RegionDropdown.tsx
```diff
@@ -0,0 +1,51 @@
...
+  );
+
+  const handleChange = (id: DhbRegionId) => {
+    setRegion(id);
+    onChange(id);
+  };
+
+  useEffect(() => {
+    onChange(region);
+  }, []);
```
andycarrell, line 34: Is there a particular reason to sync state rather than lift state up a level?

app/pages/T7cn0yCTBUmUlBfQq3JfKA.tsx
```diff
@@ -0,0 +1,146 @@
...
+    aucklandDhbsPopulationDoseData.sort((a, b) =>
+      a.dhbName.localeCompare(b.dhbName)
+    );
+
+  const hasMounted = useHasMounted();
+
+  return (
+    <div className="h-full min-h-screen">
+      <Head>
+        <meta name="robots" content="noindex" />
```
andycarrell, line 38: I think we might want?

```suggestion
        <meta name="robots" content="noindex, nofollow" />
        <meta name="googlebot" content="noindex, nofollow" />
```

app/components/RegionDropdown/RegionDropdown.tsx
jishaal: Mainly because we're persisting state in `local-storage` and I'm delaying the render until its mounted on client side.

The idea was the parent will wait for this "initial state" before attempting to show/filter the data. If we lift up then would have to delay access to initial state somewhere.

Def open to better solutions

app/pages/T7cn0yCTBUmUlBfQq3JfKA.tsx
jishaal: From what I read `googlebot` only blocks google whereas `robots` blocks most/all including Google. `nofollow` is about the crawler not following the links on the page the go to other pages. So may be fine? I can update if we wanna be safe

https://developers.google.com/search/docs/advanced/crawling/block-indexing

app/components/RegionDropdown/RegionDropdown.tsx
andycarrell: Ah gotcha - what's the implication for server rendering? We can't show anything till on the client?

jishaal: Pretty much 😬 

Or we forego the persistence for now and default Auckland regardless to keep it simple for now

andycarrell: 🤔 Yeh interesting problem - I think removing persistence might be a good move. Created it as a diff ticket so we could iterate on it

app/pages/T7cn0yCTBUmUlBfQq3JfKA.tsx
jishaal: Made the change any, this is temporary

andycarrell: Ah gotcha - I copied and pasted from one I new haha

Feel free to commit or ignore!

andycarrell (approved):

finnclark (approved):

PR #59, opened 2021-10-30
Refactor styles - allowing us to import files where needed

andycarrell:
## Motivation

Since we're using css modules, we don't _really_ need so many different "main" css files. This PR removes those files and makes them more generic.

## Changes

1. Extracts animations to their own file. Not sure if these should just live with Progress bar?
2. Moves theming (light & dark) to globals
3. Moves font import to globals
4. `main.scss` - this is pretty much all progress bar stuff now will move in next PR
5. Renames styles -> typography (contains all heading & text classnames)
6. Creates a vars file to have basic variables radius for now - could have spacing in here too?
7. Updates all imports to come through globals

### DIFF
--- a/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
+++ b/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
@@ -1,3 +1,5 @@
+@import "../../styles/vars.scss";
+
 .description-definition {
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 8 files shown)


jishaal (approved):

finnclark:
How come you are keen to put the colour theming in globals rather than keep it in colors?

andycarrell:
@finnclark - we can't import colors in other files if the theming is in there too - it's kind of a trade off I guess

PR #60, opened 2021-10-30
Refactor progress bar - css modules, and removes old css

andycarrell:
Also updates API to expect values in a decimal version (matches data store), also adds configurable goal (but set to default 0.9)


app/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,53 @@
...
+  secondDose: number;
+  size?: "small" | "large";
+  goal?: number;
+};
+
+export const Progress = ({
+  firstDose,
+  secondDose,
+  size = "small",
+  goal = 0.9,
```
andycarrell, line 16: I've updated these props to expect values less than 1 - so that they match the data store. IMO the less we manipulate on the way through the better!

app/components/Progress/Progress.module.scss
```diff
@@ -1,21 +1,45 @@
...
 .progress-bar-large {
   border-radius: $radius-large 0 0 $radius-large;
 }
 
-.progress-total {
-  background: var(--progress--total);
-}
-
 .progress-first {
+  height: 100%;
```
andycarrell, line 42: Additional css is migrated from tailwindcss

finnclark (approved):

andycarrell:
@jishaal going to send this in before any more changes conflict 😬

PR #61, opened 2021-10-30
Adds background color transition for body colors

andycarrell:
@finnclark a little something to make the switch a little bit less jarring. All good if not a fan, we can close!

### DIFF
--- a/app/styles/typography.scss
+++ b/app/styles/typography.scss
@@ -47,11 +47,11 @@
   color: var(--text-secondary);
 }
 
+.data-text-complete {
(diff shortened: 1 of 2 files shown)


app/styles/typography.scss
```diff
@@ -47,11 +47,11 @@
   color: var(--text-secondary);
 }
 
+.data-text-complete {
+  color: var(--text-positive);
+}
+
```
andycarrell, line 53: Simple move style to be alongside "data-text"

finnclark (approved):

PR #62, opened 2021-10-30
Added Fetchy Fetch Boi

theramis:
### Changes 
1. Created a Fetchy Fetch Boi with does the following 
  a. Gets the data from the MOH site 
  b. Check if data for that day is already stored in the repo. If store it. 
  c. If yes, then update it if the new data is different. 
2. Made a shared folder with some stuff that would be used by both the app and fetchy boi
3. Added workflow to build fetchy 

### Questions 
1. ~~Is ts-node the right way to actually run the script?~~ Yes
2. ~~Is tsc the right way to build for CI, to test fetchy hasn't been broken?~~ 
    a. ~~If yes whats the best way to ignore the *.js files it generates. I'm assuming gitignore for all js files in the fetchyfetchboi folder?~~ avoiding this problem by just running the app in the build
3. ~~The tsconfig file is currently inside the app folder. I didn't have any issues but do we also want at the root?~~ no tsc so don't need to solve this right now

### TODO in Next PRs
- Add workflow to run fetchy fetch boi on a schedule
- Update above workflow to commit changes
- At some point link data to app
- Maybe convert app to be SSG

### DIFF
--- a/fetchyFetchBoi/extractors/dailyUpdatedTimeExtractor.ts
+++ b/fetchyFetchBoi/extractors/dailyUpdatedTimeExtractor.ts
@@ -1,0 +1,31 @@
+import { DateTime } from "luxon";
+import { convertToNumber } from "../utilities";
+
+export const extractDailyUpdatedNzTime = (rawHtml: string): DateTime => {
+  // extracts a date like "11:59pm 24 October 2021"
+  const match =
+    /Data\sin\sthis\ssection\sis\sas\sat\s([0-9]+):([0-9]+)(am|pm)\s([0-9]+)\s(.*?)\s([0-9]+)\sand\sis\supdated\sdaily/g.exec(
+      rawHtml
+    )!;
+
+  const hour = convertToNumber(match[1]);
+  const minute = convertToNumber(match[2]);
+  const isPm = match[3] === "pm";
+  const date = convertToNumber(match[4]);
+  const month = getMonthFromString(match[5]);
+  const year = convertToNumber(match[6]);
+
+  return DateTime.fromObject(
+    {
+      day: date,
+      year: year,
+      month: month,
+      hour: isPm ? hour + 12 : hour,
+      minute: minute,
+    },
+    { zone: "Pacific/Auckland" }
+  );
+};
+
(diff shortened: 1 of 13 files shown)


shared/types/VaccineDataTypes.ts
```diff
@@ -0,0 +1,17 @@
+export type DhbVaccineDoseData = {
+  dhbName: string;
+  firstDoses: number;
+  firstDosesPercentage: number;
+  firstDosesTo90Percent: number;
```
andycarrell, line 5: Can we still have `hasMetFirstDosesTarget` etc?

theramis: I wanted to keep the data as pure as possible since this is what will get saved into the txt file. 
I reckon we could had the `hasMetFirstDosesTarget` on a new type in the app which extends this.

We are event sourcing this haha the `hasMetFirstDosesTarget` is the current interpretation :P

fetchyFetchBoi/extractors/dailyUpdatedTimeExtractor.ts
```diff
@@ -0,0 +1,31 @@
...
+      month: month,
+      hour: isPm ? hour + 12 : hour,
+      minute: minute,
+    },
+    { zone: "Pacific/Auckland" }
+  );
+};
+
+const getMonthFromString = (s: string) =>
+  new Date(Date.parse(s + " 1, 2012")).getMonth() + 1;
```
andycarrell, line 31:
```suggestion
  new Date(Date.parse(`${s} 1, 2012`)).getMonth() + 1;
```

andycarrell: Would a month map be more readable?
const months: { Jan: 1 ... } etc

shared/types/VaccineDataTypes.ts
andycarrell: Yup that makes sense!

andycarrell: Would be pretty nice to do in `getServerSideProps` etc

andycarrell (approved):

PR #63, opened 2021-10-30
Modify to filter list based on selected region

jishaal:
- Connected up the region dropdown to the data to filter
- I think the data types/models can be cleaned up further, remember being a discussion for it but I've left touching it too much right now

### DIFF
--- a/app/pages/angelfish.tsx
+++ b/app/pages/angelfish.tsx
@@ -4,29 +4,21 @@ import Head from "next/head";
 import { Progress } from "../components/Progress";
 import { DosesDescriptionList } from "../components/DosesDescriptionList";
 import { DarkModeToggle } from "../components/DarkModeToggle";
-import { DhbPopulationDoseData } from "../types/VaccineDataTypes";
-import fetchHomePageProps from "../services/homePagePropsService";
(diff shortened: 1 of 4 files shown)


theramis:
> I think the data types/models can be cleaned up further, remember being a discussion for it but I've left touching it too much right now

Yea don't worry about this too much yet. They'll probably change a little once the data from fetchy is being used haha

app/pages/angelfish.tsx
```diff
@@ -4,29 +4,21 @@ import Head from "next/head";
...
   dataFetchedAtTimeUtc: string;
 };
 
 const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
-  const { aucklandDhbsPopulationDoseData } = props;
-
-  const sortedAucklandDhbsPopulationDoseData =
-    aucklandDhbsPopulationDoseData.sort((a, b) =>
-      a.dhbName.localeCompare(b.dhbName)
-    );
```
andycarrell, line 28: 🙋  Do we want to keep alpha sort?

jishaal: Remember @finnclark mentioning the intent was to be North => South direction

andycarrell (approved):

PR #64, opened 2021-10-30
adjusts colors and removes bg shape

finnclark:
- adjusts icon button colours
- removes unneeded colour vars
- adjusts some bg colours
- removes background shape


andycarrell (approved):
@theramis 👋 rip the big U

PR #65, opened 2021-10-30
Adds Waitematā, correctly spelled.

andycarrell:
@finnclark this also sets up a map from dhb name to whatever we want - I've only changed Waitematā so far but keen for more suggestions


app/utilities/dhbDisplayName.ts
```diff
@@ -0,0 +1,37 @@
...
+  Tairawhiti: "Tairawhiti",
+  Taranaki: "Taranaki",
+  Waikato: "Waikato",
+  Wairarapa: "Wairarapa",
+  Waitemata: "Waitematā",
+  "West Coast": "West Coast",
+};
+
+const dhbDisplayName = (raw: string): string => {
+  console.log(raw, nameMap[raw], raw in nameMap);
```
theramis, line 29: Did you want to leave this in?

app/utilities/dhbDisplayName.ts
```diff
@@ -0,0 +1,37 @@
...
+  "Bay of Plenty": "Bay of Plenty",
+  Canterbury: "Canterbury",
+  "Capital and Coast": "Capital and Coast",
+  "Counties Manukau": "Counties Manukau",
+  "Hawkes Bay": "Hawkes Bay",
+  "Hutt Valley": "Hutt Valley",
+  Lakes: "Lakes",
+  MidCentral: "MidCentral",
+  "Nelson Malborough": "Nelson Malborough",
+  "Overseas / Unkown": "Overseas / Unkown",
```
theramis, line 16:
```suggestion
  "Overseas / Unknown": "Overseas / Unknown",
```

theramis (approved):
Smol changes, merge whenever you are happy.

app/utilities/dhbDisplayName.ts
andycarrell: NO!

console.log considered harmful

finnclark (approved):
Heck yeah!

app/utilities/dhbDisplayName.ts
```diff
@@ -0,0 +1,36 @@
+type NameMapType = {
+  [key: string]: string;
+};
+
+const nameMap: NameMapType = {
```
jishaal, line 5: Can do `const nameMap: Record<string, string>`

Ideally once all this is merged we create a type so we are consistent with the serverside name, so would be something like 

`const nameMap: Record<DhbName, string>`

andycarrell: Liiiiit exactly what I'm looking for

jishaal (approved):

PR #68, opened 2021-10-30
officially elon musk by how much space ive conquered

finnclark:
- adds sass map
- defines spacing vals

### DIFF
--- a/app/styles/spacing.scss
+++ b/app/styles/spacing.scss
@@ -1,0 +1,62 @@
+// Spacing definitions
+
+$space-2xs: .25rem;   // 4px
+$space-xs: .5rem;     // 8px
+$space-s: .75rem;     // 12px
+$space-m: 1rem;       // 16px
+$space-l: 1.25rem;    // 20px 
+$space-xl: 1.5rem;    // 24px
+$space-2xl: 2rem;     // 32px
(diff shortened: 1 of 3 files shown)


andycarrell (approved):
This is LIT!

Suggestions (don't block this PR tho)
- Add `0`
- Add height and width to the each
- I have a minor preference for x & y instead of vertical & horizontal

In a future PR, it'll be really helpful to have "space" and "gap" utilities which can use the same scale

app/styles/spacing.scss
```diff
@@ -0,0 +1,72 @@
...
+
+$space-2xs: 0.25rem; // 4px
+$space-xs: 0.5rem; // 8px
+$space-s: 0.75rem; // 12px
+$space-m: 1rem; // 16px
+$space-l: 1.25rem; // 20px
+$space-xl: 1.5rem; // 24px
+$space-2xl: 2rem; // 32px
+$space-3xl: 2.5rem; // 40px
+$space-4xl: 3rem; // 48px
```
andycarrell, line 11: We miiiight need to move this into vars actually if we want to use them in module.scss files?

finnclark: can do

PR #70, opened 2021-10-30
Reverts deprecated homepage to tailwind

andycarrell:
This change allows us to iterate on /angelfish without breaking the homepage - specifically for the font sizes / line heights, and removing the check box from the detail list. 

Note that this is going to look pretty confusing from a classNames PoV - if the intention is we don't make anymore changes to this page, I think it's fine 🙏

### DIFF
--- a/app/pages/index.tsx
+++ b/app/pages/index.tsx
@@ -1,8 +1,7 @@
 import { GetServerSideProps } from "next";
-import {
-  DosesDescription,
-  DosesDescriptionList,
-} from "../components/DosesDescriptionList";
+import cx from "classnames";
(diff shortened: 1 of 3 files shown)


app/pages/index.tsx
```diff
@@ -134,6 +151,61 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
 
 export default Home;
 
+export const DosesDescription = ({
```
andycarrell, line 154: Copy from description list and refactor - this page will be removed eventually anyway 🙏

jishaal (approved):

finnclark (approved):
i dont know what is happening here but go for it lol

PR #71, opened 2021-10-30
Adds space x & y - uses in doses description list

andycarrell:
Adds `space-x-` - still not sure on naming :( `gap` seems perfect, but gap is already a css property (for grid) and I don't want to overload that meaning. Some other chandidates:

- `child-space-` or `child-gap-`
- `slot-`
- `spread-`

I kind of like "spread" actually

### DIFF
--- a/app/styles/spacing.scss
+++ b/app/styles/spacing.scss
@@ -72,4 +72,16 @@ $space-map: (
   .width-#{$size} {
     width: $space;
   }
+
+  // Space
+  .space-y-#{$size} {
+    & > :not([hidden]) ~ :not([hidden]) {
+      margin-top: $space;
+    }
+  }
+  .space-x-#{$size} {
+    & > :not([hidden]) ~ :not([hidden]) {
+      margin-left: $space;
+    }
+  }
 }
--- a/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
+++ b/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
@@ -34,7 +34,6 @@
   color: var(--progress-complete);
   background: var(--complete-background);
   border-radius: $radius-full;
-  margin-right: 0.5rem;
 
   & svg {
     width: 0.875rem;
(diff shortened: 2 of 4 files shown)


app/styles/spacing.scss
```diff
@@ -72,4 +72,16 @@ $space-map: (
...
   }
+
+  // Space
+  .space-y-#{$size} {
+    & > :not([hidden]) ~ :not([hidden]) {
+      margin-top: $space;
+    }
+  }
+  .space-x-#{$size} {
+    & > :not([hidden]) ~ :not([hidden]) {
```
andycarrell, line 83: this selector selects all subsequent siblings of a child element that is not hidden, (given that the siblings are also not hidden). 

What this practically means, is that the first child doesn't get the margin - hence why it's margin-top for y and margin-left for x 🙏

jishaal: Interesting hadn't seen this as a solution before! Nice

jishaal (approved):

app/styles/spacing.scss
andycarrell: Guess where I stole it from ;)

andycarrell:
@jishaal - what do you think about the name "spread-" ie `spread-x-xl`

Or are you a fan of space- ??

jishaal:
@andycarrell I think space is more clear + less cognitive load to understand what it means

andycarrell:
Let's do this then! cc @finnclark if you're happy with merge :D

app/styles/spacing.scss
finnclark: I don't understand this even with your explanation haha

app/styles/spacing.scss
```diff
@@ -72,4 +72,16 @@ $space-map: (
...
+      margin-left: $space;
```
finnclark, line 84: is this supposed to be margin left? doesn't seem to make sense with "x"? I thought x would be both left and right?

finnclark (approved):
ngl the hidden thing seems overly complicated but maybe it's just too smart for my tiny brain. 

Can't you just use the margin classes on the children?

app/styles/spacing.scss
```diff
@@ -72,4 +72,16 @@ $space-map: (
   .width-#{$size} {
     width: $space;
   }
+
+  // Space
+  .space-y-#{$size} {
+    & > :not([hidden]) ~ :not([hidden]) {
+      margin-top: $space;
```
finnclark, line 79: Same with this - is this supposed to be top and bottom?

PR #72, opened 2021-10-30
Adds flex scss & uses where appropriate

andycarrell:
We're gonna need flex, without tailwind. This PR adds it in as a scss sheet - it's very lightweight for now, but can be added to in the future as more properties are needed 🙏

### DIFF
--- a/app/pages/angelfish.tsx
+++ b/app/pages/angelfish.tsx
@@ -69,17 +69,15 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
         </div>
       </section>
       <footer className="pb-4 pt-20 md:pt-32">
-        <div className="page-container mx-5 lg:mx-auto">
-          <div className="flex flex-row align-items-center justify-content-center footnote">
-            <p>
-              Data source:{" "}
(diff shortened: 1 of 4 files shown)


app/styles/flex.scss
```diff
@@ -0,0 +1,43 @@
+.flex {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  align-items: start;
+}
```
andycarrell, line 6: Note the default of column & start & start

IMO this is nice, because it matches a block layout.

Can easily be changed tho 🕺

app/pages/angelfish.tsx
```diff
@@ -32,7 +32,7 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
       </Head>
       <section className="pb-12 pt-12 md:pt-16">
         <div className="page-container mx-5 lg:mx-auto">
-          <div className="flex flex-row justify-between">
+          <div className="flex flex-row justify-content-between">
```
andycarrell, line 35: Note change in convention from tailwind. I like the tailwind naming, but this is nice too imo

jishaal (approved):

PR #73, opened 2021-10-30
Adds workflow_dispatch for manual fetch

andycarrell:
@theramis


theramis:
You're gonna have to change the if statement at the bottom of the workflow. To also commit changes for workflow dispatch events

theramis (changes requested):
Need to update if statement for workflow dispatch

theramis (approved):

PR #74, opened 2021-10-30
Adds a resets scss file - imports in globals

andycarrell:
Tailwind applies css resets to remove default browser styling - when we remove tailwind, we'll need to remove the resets too 🙏 I took a look at their resets and pulled some of the ones we'll need off the bat.

Question: This is pretty common (we do it at work etc) - are there any resets I've missed / any you wouldn't have in here?

### DIFF
--- a/app/styles/globals.scss
+++ b/app/styles/globals.scss
@@ -2,6 +2,8 @@
 @import "tailwindcss/components";
 @import "tailwindcss/utilities";
 
+@import "resets";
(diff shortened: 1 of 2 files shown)


finnclark (approved):

jishaal:
I've generally used normalise.css, though not sure what the recommendations for this is in 2021

We can build upon if we find more things needing resetting

PR #75, opened 2021-10-31
Breakpoints and removes Tailwind from /angelfish

andycarrell:


app/pages/angelfish.tsx
```diff
@@ -25,61 +25,66 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
...
               Ministry of Health NZ
             </ExternalLink>
           </p>
-          <div className="flex flex-row items-center justify-center footnote mt-1">
-            <div className="mr-1">&#128075;</div>
+          <div className="flex flex-row items-center justify-center footnote margin-top-2xs">
+            <div className="margin-right-2xs">&#128075;</div>
             <p>
-              Made <span className="hidden sm:inline">with love</span> by{" "}
+              Made <span hidden>with love</span> by{" "}
```
andycarrell, line 87: Hidden for now. Will make a hidden component next with new breakpoints

app/pages/angelfish.tsx
```diff
@@ -25,61 +25,66 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
   const [region, selectedRegion] = useState<DhbRegionId>("auckland");
 
   return (
-    <div className="page">
+    <main className="page flex align-items-center">
```
andycarrell, line 28: Manually styling this as flex for now, instead of assuming it in the page class name

app/styles/flex.scss
```diff
@@ -2,7 +2,7 @@
   display: flex;
   flex-direction: column;
   justify-content: flex-start;
-  align-items: start;
+  align-items: flex-start;
```
andycarrell, line 5: Missed on first pass

app/styles/vars.scss
```diff
@@ -1,3 +1,10 @@
+// Breakpoints
+
+$breakpoint-xs: 360px;
+$breakpoint-s: 640px;
+$breakpoint-m: 768px;
+$breakpoint-l: 1024px;
```
andycarrell, line 6: @jishaal let me know what you think - it's basically same as tailwind, with a cheeky xs breakpoint just in case we need a slightly different design on very very small pages

jishaal: Look good to me, don't imagine us need any larger rn

jishaal (approved):

app/styles/vars.scss
finnclark, line 6:
```suggestion
$breakpoint-xs: 360px;
$breakpoint-s: 640px;
$breakpoint-m: 800px;
$breakpoint-l: 1000px;
```

Thoughts on making these more rounded? the specific values are pretty arbitrary, so personal preference is to make them a little easier to remember

andycarrell: I think it's probably that 1024 fits into the "harmonic" scale - for example 64 goes 8 times into 1024.

I would say as well it will make the container width of 1000px a little bit more difficult to fit into a breakpoint of 1000px hahah

But other than, no reason not to! Would suggest we maybe we drop the md breakpoint down to 750?

finnclark: Sweet as let's just go with as is and change later if we need :)

PR #76, opened 2021-10-31
Region dropdown styles

jishaal:
Opening as draft for review + so I can test different devices with the preview build


app/components/RegionDropdown/RegionDropdown.module.scss
```diff
@@ -1,53 +1,45 @@
...
-  @include h2;
   font-weight: 500;
   width: 100%;
   padding: 1rem 1.5rem 1rem 2rem;
   border-radius: 8px;
   cursor: pointer;
+  color: inherit;
 
   &:focus {
     background: rgba(255, 255, 255, 0.05);
```
andycarrell, line 34: Can we drop default outline etc here?

jishaal: Yeah we can, but do we have an alternative style, especially for light version

jishaal: cc @finnclark

andycarrell: Could do a simple "ring" using tailwind for now, use ring-current so it matches text?

https://tailwindcss.com/docs/ring-color

finnclark: Should we avoid adding tailwind stuff if we're moving away from it @andycarrell ?

andycarrell: As something to unblock this PR and be ready to ship I think it's fine? Also, we can then copy that styling if we're happy with it

finnclark (approved):
I'm happy to approve this - all good if I fiddle with styles in a new PR to suss the hover/focus + padding values etc afterwards?

jishaal:
Yes please do you CSS magic 🙏🏼 🧙🏼 @finnclark

PR #77, opened 2021-10-31
Allowing fetchy to be trigger by a repo dispatch event

theramis:
Adding in this new event for when we want to create an endpoint in netlify to trigger this workflow
image

Also removed the push event since we don't really need it to run once its been merged into the main branch. It would have gone through a PR already. (also when I use the github API to check when the last run was, I don't want the push events included so this would remove them)

P.S. Promise this is the last rename of the fetchy workflow

### DIFF
--- a/.github/workflows/fetchy.yml
+++ b/.github/workflows/fetchy.yml
@@ -1,19 +1,18 @@
 name: Run Fetchy Fetch Boi
 on:
+  # will be removed soon
   schedule:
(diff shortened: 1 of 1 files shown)


andycarrell:
Question: why do we need a repository dispatch? Can't the api endpoint call the function directly?

andycarrell (approved):

jishaal (approved):

PR #78, opened 2021-10-31
Update meta tags + theme color for Angelfish release

jishaal:
Draft PR to tidy meta tags for release + get feedback

### Changes
- Update title + description to better reflect new changes of all DHB data
- Update theme color to match new background
- Add twitter + facebook open graph tags. Need an image to the image preview though it should still work without one (be blank)


app/pages/_app.tsx
```diff
@@ -4,34 +4,34 @@ import Head from "next/head";
 import "../styles/globals.scss";
 
 const metaTags = {
-  title: "Auckland DHB Vaccination Rates to 90%",
+  title: "Aotearoa NZ Vaccination Rates to 90% - Outoflockdown.co",
```
jishaal, line 7: Not sure how we should word the title + description. Appreciate any feedback

andycarrell: I like it one small suggestion:

```suggestion
  title: "Aotearoa NZ Vaccination Rates to 90% - outoflockdown.co",
```

app/pages/_app.tsx
```diff
@@ -4,34 +4,34 @@ import Head from "next/head";
...
   description:
-    "Easily view Auckland's vaccinations to 90% by DHB and the daily progress towards getting out of lockdown.",
+    "Easily view vaccinations to 90% by DHB for Auckland, the rest of New Zealand and the daily progress towards getting out of lockdown.",
```
finnclark, line 9:
```suggestion
    "View vaccination rates for each District Health Board in New Zealand and track the daily progress towards the 90% goal, enabling Kiwis to get out of lockdown.",
```

app/pages/_app.tsx
finnclark: Maybe `New Zealand` instead of `NZ` too?

app/pages/_app.tsx
jishaal: Love it! Ta

jishaal:
Even without the card image, we now get this rendering in Twitter as opposed to nothing

image

https://cards-dev.twitter.com/validator

app/pages/_app.tsx
```diff
@@ -4,34 +4,34 @@ import Head from "next/head";
 import "../styles/globals.scss";
 
 const metaTags = {
-  title: "Auckland DHB Vaccination Rates to 90%",
+  title: "Aotearoa New Zealand Vaccination Rates to 90% - outoflockdown.co",
   description:
-    "Easily view Auckland's vaccinations to 90% by DHB and the daily progress towards getting out of lockdown.",
+    "View vaccination rates for each District Health Board in New Zealand and track the daily progress towards the 90% goal, enabling Kiwis to get out of lockdown.",
```
finnclark, line 9: Sorry minor wording tweak again

```suggestion
    "View vaccination rates for each District Health Board (DHB) in New Zealand and track daily progress towards the 90% goal, enabling Kiwis to get out of lockdown.",
```

finnclark (approved):

app/pages/_app.tsx
```diff
@@ -4,34 +4,34 @@ import Head from "next/head";
 import "../styles/globals.scss";
 
 const metaTags = {
-  title: "Auckland DHB Vaccination Rates to 90%",
+  title: "Aotearoa New Zealand Vaccination Rates to 90% - outoflockdown.co",
   description:
-    "Easily view Auckland's vaccinations to 90% by DHB and the daily progress towards getting out of lockdown.",
+    "View vaccination rates for each District Health Board (DHB) in New Zealand and track daily progress towards the 90% goal, enabling Kiwis to get out of lockdown.",
+  url: "https://outoflockdown.co",
+  imageUrl: "https://outoflockdown.co/preview-cover.png",
```
andycarrell, line 11: Might be worth putting a env var on this - not necessary for now

finnclark (approved):

app/pages/_app.tsx
andycarrell: Actually if we just drop the uri:

```suggestion
  imageUrl: "/preview-cover.png",
```

andycarrell (approved):

app/pages/_app.tsx
jishaal: Will have to be env var based imo may leave for now. Relative won't work since it's not a resource the user/client resolves when they land on the page but rather something the site that the link being shared on tries to resolve

https://stackoverflow.com/questions/9858577/open-graph-can-resolve-relative-url/9858694#9858694

andycarrell: Ah gotcha - That's definitely my bad 🙏 

Let's merge as is, we can add an env var when we're actually battling to debug some stuff

PR #79, opened 2021-10-31
Add endpoint for triggering fetchy boi

theramis:
## Changes
- Added a new endpoint which we can use to trigger fetchy boi
- Moved .env.local file inside the app folder as it wasn't being used at the root level 

If you make a call to `/api/trigger-fetchy-boi` you can trigger the github build. (this feels super insecure lol but yolo)


andycarrell:
Any reason not to put it behind a simple token at least?

theramis:
> Any reason not to put it behind a simple token at least?

No reason, I can add some in the next PR if ya don't mind?

andycarrell (approved):

PR #80, opened 2021-10-31
Send an alert in discord if Fetchy fails to fetch

theramis:
If there is an exception when running fetchy this will send an alert to discord. 

Please merge https://github.com/theramis/aucvid/pull/79 as it will cause merge conflicts.


andycarrell (approved):

fetchyFetchBoi/fetchyFetchBoi.ts
```diff
@@ -4,18 +4,35 @@ import {
...
+    } else {
+      await updateDataIfDifferentForDate(
+        vaccineData.dataValidAsAtNzTimeIso,
+        vaccineData
+      );
+    }
+  } catch (error) {
+    console.error(error);
+
+    await got.post(process.env.DISCORD_WEBHOOK!, {
```
andycarrell, line 27: Is this `!` expected to be here?

theramis: yea 😅 

the environment var is of type `string | null` so it complains. 
This seemed like the easiest fix 😅

andycarrell: ofc makes sense - I guess the other way would be to type check and not call the endpoint but yeh all gooooood

PR #81, opened 2021-10-31
Dropdown styling changes

finnclark:
- Adds control variables and mixins
- Replaces css icon with svg (sorry @jishaal I know you spent a lot of time on it)
- Tidies up padding/colours in dropdown


app/components/RegionDropdown/RegionDropdown.module.scss
```diff
@@ -1,75 +1,59 @@
...
+// Focus-ring (move somewhere more global later)
+@mixin focus-ring {
+  box-shadow: 0 0 0 2px var(--control-focus-ring);
+}
 
 .dropdown-container {
   position: relative;
   cursor: pointer;
-  width: 16.25rem;
+  width: calc($space-3xl * 7);
```
finnclark, line 13: I don't particularly like how this is a fixed width, but I couldn't for the life of me figure out how to make the width dynamic based on the trigger text width.

jishaal:
> Replaces css icon with svg (sorry @jishaal I know you spent a lot of time on it)

@finnclark yo all good. I was just using as a means to play around 😉

app/components/RegionDropdown/RegionDropdown.module.scss
jishaal: AFAIK its not possible with native select, will need to move to a custom dropdown for that behaviour

app/pages/angelfish.tsx
```diff
@@ -30,14 +30,14 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
         <meta name="robots" content="noindex, nofollow" />
         <meta name="googlebot" content="noindex, nofollow" />
       </Head>
-      <section className="pb-12 pt-12 md:pt-16">
+      <section className="margin-top-2xl margin-bottom-xl md:pt-16">
```
finnclark, line 33: I'll do the rest of these in a seperate PR

app/components/RegionDropdown/RegionDropdown.module.scss
```diff
@@ -1,75 +1,59 @@
...
   color: var(--text-positive);
 
-  //Animation is currently off
+  // Animation is currently off — I don't know if this is what we want anyways because the dropdown could be closed but still in focus?
```
jishaal, line 16: Yeah most likely only have it animated when the dropdown is open instead of on focus

app/components/RegionDropdown/RegionDropdown.module.scss
andycarrell: Why is this a calc tho? Should it be part of the spacing scale?

app/components/RegionDropdown/RegionDropdown.module.scss
finnclark: IDK how to do that though :(

jishaal: Ah, I don't think its possible with this version. Can do it in v2 ;)

app/components/RegionDropdown/RegionDropdown.module.scss
finnclark: It was my attempt at adding _some_ meaning to the value... 😅 

I don't think it should be part of the spacing scale because it's pretty bespoke and not really used in the same context as the other spacing values. We could add like "standard control size" variables or something?

andycarrell: Na I reckon just hard code it here - if we come to creating a standard set we can loop back round?

Could always make a local variable so it's actually named which gives it some meaning 🤔

app/pages/angelfish.tsx
andycarrell: I think it's mostly done in my PR but will need the design eye ;)

app/components/RegionDropdown/RegionDropdown.module.scss
finnclark: donezo

andycarrell (approved):

jishaal (approved):

jishaal:
NIce finn thanks for tidying this up. Could prob use the same focus ring on the dark mode button

PR #86, opened 2021-10-31
Updates grid to have region as the key, so when changed it reanimates

andycarrell:
Also small bug on the home page, width needs to be 100%

### DIFF
--- a/app/pages/angelfish.tsx
+++ b/app/pages/angelfish.tsx
@@ -48,7 +48,7 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
         </section>
         <section>
           <div className="page-container">
-            <div className="doses-data-grid">
+            <div className="doses-data-grid" key={region}>
               {allDhbsPopulationDoseData
                 .filter((dhb) => dhb.regionIds.includes(region))
                 .map((dhb) => (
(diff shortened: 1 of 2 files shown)


app/pages/angelfish.tsx
```diff
@@ -48,7 +48,7 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
         </section>
         <section>
           <div className="page-container">
-            <div className="doses-data-grid">
+            <div className="doses-data-grid" key={region}>
```
andycarrell, line 51: Forces remount so the bars re-animate - @jishaal any worries with performance of this?

app/styles/globals.scss
```diff
@@ -31,6 +31,7 @@ body {
   min-height: 100vh;
 
   & .page-container {
+    width: 100%;
```
andycarrell, line 34: v important haha max sure it does take up all of the max-width

app/pages/angelfish.tsx
jishaal: Nah should be fine imo 👌🏼

jishaal (approved):

finnclark (approved):

PR #94, opened 2021-10-31
Responsive scaling

finnclark:
- Fixes dropdown padding (was off by 4px)
- Adds rem scaling for small viewports
- Removes padding removal for headings

### DIFF
--- a/app/styles/globals.scss
+++ b/app/styles/globals.scss
@@ -18,14 +18,6 @@
   -webkit-tap-highlight-color: transparent;
 }
 
-html {
-  font-size: 14px;
-
-  @media (min-width: $breakpoint-s) {
-    font-size: 16px;
-  }
-}
-
 body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
--- a/app/styles/globals.scss
+++ b/app/styles/globals.scss
@@ -18,6 +18,14 @@
   -webkit-tap-highlight-color: transparent;
 }
 
+html {
+  font-size: 16px;
+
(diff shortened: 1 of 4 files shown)


app/styles/globals.scss
```diff
@@ -18,6 +18,14 @@
   -webkit-tap-highlight-color: transparent;
 }
 
+html {
+  font-size: 14px;
```
jishaal, line 22: I feel like going this way, the smallest text i.e. dose description rows are just a tad too small? The current size seems perfect. Maybe for fonts we want to be a bit more selective?

andycarrell:
@finnclark - I was thinking about this overnight, it should be relatively simple to add the breakpoint styling to typography too - so that in this case we could do:

```jsx
 <h1 className="heading-2 xs:heading-1">Aotearoa</h1>
```

And opt into shrinking on a case by case basis - what do you think? cc @jishaal

app/styles/globals.scss
finnclark: Yeah, fair - happy to be more considered :)

finnclark:
@andycarrell I would really prefer to do this globally in CSS, rather than case by case on the HTML element - means we can control all font sizing from one place. Keen to make it more granular - was just trying to get something in that didn't fiddle with the padding to solve the space issue

andycarrell:
@finnclark - Yeee that's what we do at work! 

In that case the solution is probably to go through case by case in the typography and define a small screen font-size?

We could just do that quickly on `heading-1` for a start so we can revert the padding change?

andycarrell (approved):

PR #96, opened 2021-10-31
Page components

andycarrell:
@finnclark's suggestion was the right way after thinking a little more on it 😅

### DIFF
--- a/app/components/Page/Page.module.scss
+++ b/app/components/Page/Page.module.scss
@@ -1,0 +1,21 @@
+@import "../../styles/vars.scss";
+
+$container-width: 1000px;
+
+.page {
+  height: 100%;
+  min-height: 100vh;
+}
+
+.page-container {
+  width: 100%;
+  max-width: calc($container-width + 2 * $space-m);
+
+  padding-left: $space-m;
+  padding-right: $space-m;
+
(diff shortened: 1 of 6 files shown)


app/components/Page/Page.module.scss
```diff
@@ -0,0 +1,21 @@
...
+$container-width: 1000px;
+
+.page {
+  height: 100%;
+  min-height: 100vh;
+}
+
+.page-container {
+  width: 100%;
+  max-width: calc($container-width + 2 * $space-m);
```
andycarrell, line 12: Account for padding in max width

app/pages/angelfish.tsx
```diff
@@ -1,15 +1,16 @@
...
 const Home: React.FC = () => {
   return (
-    <main className="page flex align-items-center">
+    <Page className="flex align-items-center">
       <Head>
         <meta name="robots" content="noindex, nofollow" />
         <meta name="googlebot" content="noindex, nofollow" />
       </Head>
-      <div className="flex-1 w-full" />
+      <div className="flex-1 width-full" />
```
andycarrell, line 13: 😬

jishaal (approved):
Nice LGTM!

PR #110, opened 2021-11-01
Ops/split actions

andycarrell:
@theramis LMK what you think - if you wanna I'll remove the git on push config


.github/workflows/fetchy.yml
```diff
@@ -30,7 +32,6 @@ jobs:
       - name: Run Fetchy
         run: npm run fetch-boi
       - uses: EndBug/add-and-commit@v7
-        if: github.event_name == 'workflow_dispatch' || github.event_name == 'repository_dispatch'
```
andycarrell, line 33: We technically wouldn't need this 👋

.github/workflows/fetchy.yml
```diff
@@ -1,10 +1,12 @@
 name: Run Fetchy Fetch Boi
 on:
-  pull_request:
   repository_dispatch:
   workflow_dispatch:
     branches:
       - main
+  push:
+    branches:
+      - ops/split-actions
```
andycarrell, line 9: This is kinda how you can test it per PR / branch

.github/workflows/verify-fetchy.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/setup-node@v2.4.1
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: "npm"
+      - name: Install dependencies
+        run: npm ci
+      - name: Run Fetchy
+        run: npm run fetch-boi
+      - name: Print results
+        run: git ls-files --others --exclude-standard | xargs -I filename cat filename
```
andycarrell, line 29: There's probably a better way ;)

theramis: TBH this isn't getting printed out at the moment so we don't need this. 

But if you want you can just do `git status`. It should be enough to tell us a new file got generated.

andycarrell: Interesting yeh ofc - we could purge the data folder prior to run, but like you say, probably not necessary

.github/workflows/verify-fetchy.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: Verify Run Fetchy
+    runs-on: ubuntu-latest
+    env:
+      # set this secret, you can get the value from this discord message https://discord.com/channels/658523382442360866/903771050385502239/904216952141152328
+      DISCORD_WEBHOOK: $ {{ secrets.DISCORD_WEBHOOK }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2.3.4
+        with:
+          token: ${{ secrets.PAT }}
```
theramis, line 16:
```suggestion
```

You need to delete these two lines :P

.github/workflows/verify-fetchy.yml
theramis:
```suggestion
        run: git status
```

theramis:
I think I personally still prefer having it all in one file but can see that it's to think about in two different workflows since there are less permutations. 

So happy to do this.

andycarrell:
Thanks @theramis - this might be one of the things I feel strongly about haha 🙏

theramis (approved):

theramis:
> Acccctually, there might be another way, to have it all in one file, but have a second step dependent on the first.
> 
> I can look into that tomorrow, but would rather get this in and PRs passing?

nah alg, this is fine :)

PR #111, opened 2021-11-01
Switch index page to use static site generation and to use the data from the repo

theramis:


shared/dataRepository.ts
```diff
@@ -39,3 +47,9 @@ const generateFilePath = (time: NzTimeIso) => {
   const date = DateTime.fromISO(time).toFormat("yyyy-LL-dd");
   return `${CONSTANTS.dataFolder}/${date}.txt`;
 };
+
+const getLatestVaccineDataFilePath = async () => {
+  const files = await fs.readdir(CONSTANTS.dataFolder);
+  const latestDataFile = files.sort()[files.length - 1];
```
andycarrell, line 53:
```suggestion
  const [latestDataFile] = files.sort().reverse();
```

pages/index.tsx
```diff
@@ -107,9 +97,9 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
 
 export default Home;
```
andycarrell, line 98: If you're gonna do it, you gotta do it all

```suggestion
export default Index;
```

shared/dataRepository.ts
theramis: > const [latestDataFile] = files.sort().reverse();

Is this common in JS land? Seems harder to understand IMO. 
Happy to change if this is usual pattern.

andycarrell: Yeh I think `at` is still proposal stage 3 - but some time soon hoepfully!

I can't comment on whether it's more common, it's probably preference - I like the destructuring version because to me it reads like:

```
const [latestDataFile] // take the first file out of a list of files sorted latest first
    = files.sort().reverse(); // sort files so that latest is first first
```

So another way of writing it might be:
```suggestion
  const sortFilesByName = f => [...f].sort().reverse();
  const [latestDataFile] = sortFilesByName(files);
```

No need to change tho, I'll approve first 🙏

andycarrell (approved):

shared/dataRepository.ts
theramis: Yea I might suggest `[files.length - 1];` is a bit more universal than destructing and keep it as it is :P

jishaal, line 53: Can we rely on this default `.sort`? Actually I don't know

theramis: I named the files like `2021-10-30.txt`. So as long as the sort is a standard string sort. I think we should be fine right?

jishaal: Yeah I think so too

jishaal (approved):

jishaal:
Nicely done guys  😍

PR #114, opened 2021-11-01
adds flex to title section, tweaks drodpown styling

finnclark:
Few styling changes. Was doing some stuff to test the new branching stuff not using my fork


app/components/RegionDropdown/RegionDropdown.module.scss
```diff
@@ -48,12 +48,13 @@ $control-width: 260px; // Control width local variable for now
 
 .dropdown-caret {
   position: absolute;
-  right: $space-xl;
-  top: $space-half;
+  right: $space-l;
+  top: calc($space-half);
```
andycarrell, line 52: Why does this need to be a `calc`?

pages/index.tsx
```diff
@@ -29,14 +29,14 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
   return (
     <Page className="flex align-items-center">
       <div className="flex-1 width-full">
-        <section className="padding-top-4xl padding-bottom-xl">
+        <section className="padding-top-4xl padding-bottom-l">
           <PageContainer>
```
andycarrell, line 33: I'm not sure we want to assume flex on all page containers - we can achieve the intended style with a class name 
for this specific case:
```suggestion
          <PageContainer className="flex align-items-stretch">
```

If we want flex he

pages/index.tsx
```diff
@@ -29,14 +29,14 @@ const Home: React.FC<HomePageProps> = (props: HomePageProps) => {
...
-            <div className="padding-left-l margin-bottom-l m:margin-bottom-4xl">
+            <div className="padding-left-l l:padding-left-none margin-bottom-l s:margin-bottom-4xl">
               <div className="width-full flex flex-row justify-content-between align-items-center">
                 <h2 className="heading-2">Vaccination rates</h2>
                 {hasMounted && <DarkModeToggle className="flex-0" />}
               </div>
-              <h1 className="heading-1">Aotearoa</h1>
+              <h1 className="heading-1 margin-bottom-xl">Aotearoa</h1>
```
andycarrell, line 39: Can we extend the spacing scale, rather than using additive margin to create enough space? IMO all the margin should be on the parent div here

app/components/RegionDropdown/RegionDropdown.module.scss
finnclark: Woops! Was playing around with the layout and forgot to revert this!

pages/index.tsx
finnclark: Yep, good call!

pages/index.tsx
finnclark: I was playing around with aligning the dropdown to the right and although I didn't actually do that, when we introduce sort, having this as flex will enable that

finnclark: Seems unnecessary to have an additional set of classes when we already have a class for this one instance of page-container? Can always abstract it later if we find we use page-container where we don't need flex?

andycarrell: This isn't the one instance of page container though? - there should be 3 on the home page alone (seems the footer one got dropped somewhere along the way!), and in the case we add sort they'll all have different styling - which is fine. 

I don't really see this as "additional classes" but styling this differently to the other cases of page container - it's like a variant of the component. The way we've decided to do that is to using the className, but it's the equivalent of creating props on the component. 

I think opt out is way more confusing than opt in to be honest - I like adding the additional classes (even if it looks unnecessary!) because it's really explicit about what this component does and how it's styled. If we make it flex by default, we'll end up in the situation where we do: `<PageContainer className="align-items-center" />` which kind of doesn't make sense imo (because the flex is missing)

app/components/Page/Page.module.scss
```diff
@@ -11,6 +11,8 @@ $container-width: 1000px;
   max-width: calc($container-width);
   padding-left: $space-m;
   padding-right: $space-m;
+  display: flex;
+  flex-direction: column;
```
andycarrell, line 15: Lost in force push!

```suggestion
```

andycarrell (approved):
Looks really good, let's ship it :D

PR #115, opened 2021-11-01
Applies box-sizing = border-box to all elements

andycarrell:
- Adjusts page container width so it's 1000 again
- Adds in inherit for flex

### Motivation
The root cause of https://github.com/theramis/aucvid/pull/113 was most probably the removal of tailwindcss - tailwind was apply a lot of resets and defaults, one of was `box-sizing: border-box;`

Border-box means padding is counted as part of an elements width, which I think IMO is the easiest mental model to work with - although @jishaal @finnclark keen to hear your thoughts.

Changing this, means we do need to go back to the original assumption about max-width - that it's 1000px + space on the sides - On that, I think it's pretty important we have a width in there that matches the designs, otherwise we'll probably run into a scenario where the page will break in unexpected ways, but the designs will look fine. 1000px is nice and easy to remember too!


app/styles/flex.scss
```diff
@@ -22,6 +22,7 @@
 }
 
 $justify-content-map: (
+  "inherit": "inherit",
```
andycarrell, line 25: Seeing this is also a possible value - if we ever need it

jishaal:
? The root cause of #113 was most probably the removal of tailwindcss - tailwind was apply a lot of resets and defaults, one of was box-sizing: border-box;

Ah really, I thought this was already applied when I checked the Devtools in one of the previews, but must have been tw giving it as you said 😅 

Nice thanks for the fix

finnclark (approved):

jishaal (approved):

PR #116, opened 2021-11-01
Reverts change to separate github actions - sets second job to only run after fetchy

andycarrell:


theramis:
TBH I find this more confusing haha

andycarrell:
@theramis pick your poison haha

Going to keep trying, to see if I can get it to work well. You can decide to merge or close, I think there's arguments for each!

andycarrell:
@theramis - if we're querying when fetchy last ran, then wouldn't it be better to have these actions separate anyway?

theramis:
> @theramis - if we're querying when fetchy last ran, then wouldn't it be better to have these actions separate anyway?

yup haha, It's easy enough to do it with one too but easier with them separate. 

TBH I prefer seperate than having one file

andycarrell:
The more I think about it, I agree!

PR #117, opened 2021-11-02
adjusts mobile margin for titles and makes dropdown full-w on mobile

finnclark:


app/components/RegionDropdown/RegionDropdown.module.scss
```diff
@@ -7,13 +7,16 @@
...
 }
 
-$control-width: 260px; // Control width local variable for now
+$select-width: $space-full;
 
 .dropdown-container {
   position: relative;
   cursor: pointer;
-  width: $control-width;
+  width: $select-width;
```
andycarrell, line 15: minor: I'd probably just make this 100% - it's a bit easier to understand at face value compared to translating it to 'select-width' then 'space-full', it also probably doesn't need to come from a system value 💃 

```suggestion
  width: 100%;
```

andycarrell (approved):
Looks real good :D

PR #118, opened 2021-11-02
Add DhbName type and use across app

jishaal:
### Changes
- Adds `DhbName` string literal type and use across app
- Found spelling issue with `nameMap` for `Nelson Marlborough` Dhbs and this is why I love types 😏 Wasn't causing any bugs though luckily 😌

### DIFF
--- a/app/utilities/dhbDisplayName.ts
+++ b/app/utilities/dhbDisplayName.ts
@@ -1,35 +1,10 @@
 import { DhbName } from "../../shared/types/VaccineDataTypes";
 
-const nameMap: Record<DhbName, string> = {
-  Auckland: "Auckland",
-  "Bay of Plenty": "Bay of Plenty",
-  Canterbury: "Canterbury",
-  "Capital and Coast": "Capital and Coast",
-  "Counties Manukau": "Counties Manukau",
-  "Hawkes Bay": "Hawkes Bay",
-  "Hutt Valley": "Hutt Valley",
-  Lakes: "Lakes",
-  MidCentral: "MidCentral",
-  "Nelson Marlborough": "Nelson Marlborough",
-  "Overseas / Unknown": "Overseas / Unknown",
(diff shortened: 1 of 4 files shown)


app/utilities/dhbDisplayName.ts
```diff
@@ -8,7 +10,7 @@ const nameMap: Record<string, string> = {
   "Hutt Valley": "Hutt Valley",
   Lakes: "Lakes",
   MidCentral: "MidCentral",
-  "Nelson Malborough": "Nelson Malborough",
+  "Nelson Marlborough": "Nelson Marlborough",
   "Overseas / Unknown": "Overseas / Unknown",
   Northland: "Northland",
```
jishaal, line 15: Is there any point in mapping names which don't have a translation?

shared/types/VaccineDataTypes.ts
```diff
@@ -15,3 +15,31 @@ export type VaccineData = {
   vaccinationsPerDhb: DhbVaccineDoseData[];
   dataValidAsAtNzTimeIso: NzTimeIso;
 };
+
+type NorthIslandDhbKeys =
+  | "Northland"
```
theramis, line 20: Why does this work? 
Is this a TS quirk?

The way I read this is `blank or "Northland" or etcetc`

Why is the blank allowed haha

app/utilities/dhbDisplayName.ts
theramis: I reckon kill it

theramis (approved):

shared/types/VaccineDataTypes.ts
jishaal: Not really blank, it is a TS/formatting quirk. Blank is not allowed

image

andycarrell:
Holy hekka this is cool thanks!

PR #120, opened 2021-11-02
Adds specific text style mixins

finnclark:
cc @andycarrell 

- Adds mixins for card text. Moved away from h3, h4 etc because it's not really following the structure properly anyways. 
- Swapped some styling over but realised that you're probably changing the structure quite a bit so happy to remove these changes from this PR

Could maybe even remove the h3 and h4 heading mixins if they aren't being used - Just keep h1 and h2?

### DIFF
--- a/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
+++ b/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
@@ -1,4 +1,5 @@
 @import "../../styles/vars.scss";
+@import "../../styles/typography.scss";
 
 .description-definition {
   display: flex;
@@ -8,6 +9,7 @@
 }
 
 .term {
+  @include dose-label;
   width: 7rem;
   display: flex;
   flex-direction: row;
@@ -16,6 +18,7 @@
 }
 
 .definition-primary {
+  @include percentage-label;
   width: 2rem;
   text-align: right;
 }
--- a/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
+++ b/app/components/DosesDescriptionList/DosesDescriptionList.module.scss
@@ -56,7 +56,7 @@
     line-height: 1rem; // 16px
   }
 }
-.percentage-label {
+.percentage-value {
   font-size: 1.125rem; // 18px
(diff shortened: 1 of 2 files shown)


app/components/DosesDescriptionList/DosesDescriptionList.module.scss
```diff
@@ -1,4 +1,5 @@
 @import "../../styles/vars.scss";
+@import "../../styles/typography.scss";
```
finnclark, line 2: mentioned in the description but I can remove the changes from this file if that's preferred

andycarrell: Na this is all good! I can rebase ;)

andycarrell:
Some design feedback - just an opinion tho, don't have to change 🙏 

- I think the uppercase text should have a heavier font weight (500 in screenshot)
- Should we align the text to "baseline" across the term, percentage & doses remaining?

image

finnclark:
Sorry my bad, this PR was just to add the mixins for you to rebase into your new layout PR, I'm not actually keen to update the production layout design to use these styles 😊

The font-weight change looks ok but I'd be keen to double check with the new layout when you update your PR if that's ok?

andycarrell:
@finnclark - yup that makes sense, before we merge we can remove the mixins, then it won't ship 🙏 

From an implementation perspective, these classnames are global, but they're very specific - for example `vaccine-count`, it's pretty clear where that's expected to be used, but it's not that clear how we would style the next piece of text that ultimately has the same size / height / weight - so I don't think these should be mixins.

I think there's two options here either we:

- Avoid using mixins for typography and move the specific styling to `DosesDescriptionList.module.scss`. It might also make sense to create a font-size / weight scale that DosesDescription can reference

- Make the mixins more generic, probably by bringing back a scale like we had: heading-1 heading-2 etc in typography, then each component determines which classname it should use. If we're doing that, we probably don't need a mixin since we can just create classNames for each?

finnclark:
Updated per call @andycarrell

andycarrell (approved):

PR #122, opened 2021-11-03
RFC - Adds responsive styling to flex rules

andycarrell:
I get the feeling we'll imminently need these, BUT doing it this way does generate a _lot_ of `css`

### DIFF
--- a/app/styles/flex.scss
+++ b/app/styles/flex.scss
@@ -1,25 +1,4 @@
-.flex {
-  display: flex;
-  flex-direction: column;
-  justify-content: flex-start;
-  align-items: flex-start;
-}
-
-.flex-row {
-  flex-direction: row;
-}
-
-.flex-column {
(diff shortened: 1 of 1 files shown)


app/styles/flex.scss
```diff
@@ -46,8 +19,53 @@ $align-items-map: (
   "stretch": "stretch",
 );
 
-@each $name, $property in $align-items-map {
-  .align-items-#{$name} {
-    align-items: #{$property};
+@each $prefix, $breakpoint in $breakpoint-map {
+  .#{$prefix}flex {
```
jishaal, line 23: Does this output classes such as` xsflex`, `lflex`?

andycarrell: The colon is in the prefix, so the classes are s:flex, m:flex etc

jishaal (approved):

PR #123, opened 2021-11-03
Calculate vaccine dose delta and percentage

jishaal:
### Changes 🆕 

- Added delta/dose change and render in the list
- Gone with a approach at deriving it at the ~controller~ props generator level (give us ya feedback)
- Will tidy up code a bit more tomorrow, wanted to fix some types and maybe move things around but lets see

### DIFF
--- a/app/components/DosesDescriptionList/DosesDescriptionList.tsx
+++ b/app/components/DosesDescriptionList/DosesDescriptionList.tsx
@@ -20,7 +20,7 @@ const DosesDescription = ({
   hasMetTarget,
   dosesPercent,
   children = null,
-}: DosesDescriptionProps) => (
+}: Omit<DosesDescriptionProps, "dosesChange">) => (
   <div className="width-half">
     <dt className={cx(styles["dose-label"], "margin-bottom-2xs")}>{term}</dt>
     <dd className="flex flex-row justify-content-start align-items-center space-x-xs">
@@ -49,7 +49,6 @@ export const DosesDescriptionList = ({
       term="First doses"
       hasMetTarget={dhbData.hasMetFirstDoseTarget}
       dosesPercent={dhbData.firstDosesPercentage}
-      dosesChange={dhbData.firstDosesChange}
     >
       <dd className={styles["vaccine-count"]}>
         {numberFormatter.format(dhbData.firstDosesTo90Percent)} left{" "}
@@ -60,7 +59,6 @@ export const DosesDescriptionList = ({
       term="Second doses"
       hasMetTarget={dhbData.hasMetSecondDoseTarget}
       dosesPercent={dhbData.secondDosesPercentage}
-      dosesChange={dhbData.secondDosesChange}
     >
       <dd className={styles["vaccine-count"]}>
         {numberFormatter.format(dhbData.secondDosesTo90Percent)} left{" "}
--- a/app/components/DosesDescriptionList/DosesDescriptionList.tsx
+++ b/app/components/DosesDescriptionList/DosesDescriptionList.tsx
@@ -9,12 +9,13 @@ type DosesDescriptionProps = {
   term: string;
   hasMetTarget: boolean;
   dosesPercent: number;
+  dosesChange: number;
   children?: React.ReactNode;
 };
 
 type DosesDescriptionListProps = { dhbData: DhbVaccineDoseDataForIndexPage };
 
-export const DosesDescription = ({
+const DosesDescription = ({
   term,
   hasMetTarget,
   dosesPercent,
@@ -48,18 +49,22 @@ export const DosesDescriptionList = ({
       term="First doses"
       hasMetTarget={dhbData.hasMetFirstDoseTarget}
       dosesPercent={dhbData.firstDosesPercentage}
+      dosesChange={dhbData.firstDosesChange}
     >
       <dd className={styles["vaccine-count"]}>
-        {numberFormatter.format(dhbData.firstDosesTo90Percent)} left
+        {numberFormatter.format(dhbData.firstDosesTo90Percent)} left{" "}
+        {!dhbData.hasMetFirstDoseTarget && `(${dhbData.firstDosesChange})`}
       </dd>
     </DosesDescription>
     <DosesDescription
       term="Second doses"
       hasMetTarget={dhbData.hasMetSecondDoseTarget}
       dosesPercent={dhbData.secondDosesPercentage}
+      dosesChange={dhbData.secondDosesChange}
     >
       <dd className={styles["vaccine-count"]}>
-        {numberFormatter.format(dhbData.secondDosesTo90Percent)} left
+        {numberFormatter.format(dhbData.secondDosesTo90Percent)} left{" "}
+        {!dhbData.hasMetSecondDoseTarget && `(${dhbData.secondDosesChange})`}
       </dd>
     </DosesDescription>
   </dl>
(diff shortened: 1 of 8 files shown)


app/components/DosesDescriptionList/DosesDescriptionList.tsx
```diff
@@ -50,7 +51,8 @@ export const DosesDescriptionList = ({
       dosesPercent={dhbData.firstDosesPercentage}
     >
       <dd className={styles["vaccine-count"]}>
-        {numberFormatter.format(dhbData.firstDosesTo90Percent)} left
+        {numberFormatter.format(dhbData.firstDosesTo90Percent)} left{" "}
+        {!dhbData.hasMetFirstDoseTarget && `(${dhbData.firstDosesChange})`}
```
andycarrell, line 55: IMO this should probably be formatted?

```suggestion
        {!dhbData.hasMetFirstDoseTarget && `(${numberFormatter.format(dhbData.firstDosesChange)})`}
```

app/components/DosesDescriptionList/DosesDescriptionList.tsx
```diff
@@ -59,7 +61,8 @@ export const DosesDescriptionList = ({
       dosesPercent={dhbData.secondDosesPercentage}
     >
       <dd className={styles["vaccine-count"]}>
-        {numberFormatter.format(dhbData.secondDosesTo90Percent)} left
+        {numberFormatter.format(dhbData.secondDosesTo90Percent)} left{" "}
+        {!dhbData.hasMetSecondDoseTarget && `(${dhbData.secondDosesChange})`}
```
andycarrell, line 65: Minor: Preference to read in the positive intent

```suggestion
        {dhbData.hasMetSecondDoseTarget ? null : `(${dhbData.secondDosesChange})`}
```

```suggestion
        {dhbData.hasMetSecondDoseTarget || `(${dhbData.secondDosesChange})`}
```

app/propsGenerators/indexPagePropsGenerator.ts
```diff
@@ -65,13 +68,24 @@ function generateAllDhbsVaccineDoseData(
...
     return {
       ...dhb,
       regionIds: regions,
       hasMetFirstDoseTarget:
         dhb.firstDosesPercentage >= CONSTANTS.firstDoseTargetPercentage,
       hasMetSecondDoseTarget:
         dhb.secondDosesPercentage >= CONSTANTS.secondDoseTargetPercentage,
+      firstDosesChange: yesterdayDoseData
+        ? dhb.firstDosesTo90Percent - yesterdayDoseData?.firstDosesTo90Percent
+        : 0,
```
andycarrell, line 85: ❓ - Should these default to `null` ? If we ever had a historical view, 0 wouldn't be valid for the first day 🤔 

OMG we should do a historical view 👁️

shared/dataRepository.ts
```diff
@@ -48,8 +50,11 @@ const generateFilePath = (time: NzTimeIso) => {
...
 
-const getLatestVaccineDataFilePath = async () => {
+const getVaccineDataFilePath = async (period: "latest" | "yesterday") => {
   const files = await fs.readdir(CONSTANTS.dataFolder);
-  const latestDataFile = files.sort()[files.length - 1];
-  return `${CONSTANTS.dataFolder}/${latestDataFile}`;
+  const filePath =
+    period === "latest"
+      ? files.sort()[files.length - 1]
+      : files.sort()[files.length - 2];
```
andycarrell, line 58: Minor: Surely now this works 😅 

```suggestion
  const [latestPath, yesterdayPath] = files.sort().reverse();
  const filePath = period === "latest" ? latestPath : yesterdayPath;
```

shared/dataRepository.ts
```diff
@@ -48,8 +50,11 @@ const generateFilePath = (time: NzTimeIso) => {
   return `${CONSTANTS.dataFolder}/${date}.txt`;
 };
 
-const getLatestVaccineDataFilePath = async () => {
+const getVaccineDataFilePath = async (period: "latest" | "yesterday") => {
```
andycarrell, line 53: Should this use the type??

```suggestion
const getVaccineDataFilePath = async (period: DataPeriod) => {
```

andycarrell (commented):
Awesome. Love it.

If there's design tweaks to make, we could remove displaying the number in this PR and merge it in so the data has landed?

finnclark:
> Awesome. Love it.
> 
> 
> 
> If there's design tweaks to make, we could remove displaying the number in this PR and merge it in so the data has landed?

Yeah makes sense to me!

app/propsGenerators/indexPagePropsGenerator.ts
```diff
@@ -65,13 +68,24 @@ function generateAllDhbsVaccineDoseData(
...
+      secondDosesChange: yesterdayDoseData
+        ? dhb.secondDosesTo90Percent - yesterdayDoseData?.secondDosesTo90Percent
```
andycarrell, line 87: Maybe this should come from secondDoses? Also just thinking it should be a positive value 🤔 current designs are neg symbol, but it could be positive in the future too 🤔 

```suggestion
        ? dhb.secondDoses - yesterdayDoseData?.secondDoses
```

app/components/DosesDescriptionList/DosesDescriptionList.tsx
jishaal: Removed from UI for now

app/components/DosesDescriptionList/DosesDescriptionList.tsx
jishaal: Removed from UI for now

app/propsGenerators/indexPagePropsGenerator.ts
jishaal: Made it null if there is not data for yesterday. Not sure fi have done the best null checks lol

shared/dataRepository.ts
jishaal: Lol of course!

shared/dataRepository.ts
jishaal: Yep my bad, have fixed this

app/propsGenerators/indexPagePropsGenerator.ts
jishaal: Yep good point. I did go this route but went this way instead. Changed to be be positive and leave to symbol up to the UI to render

app/components/DosesDescriptionList/DosesDescriptionList.tsx
```diff
@@ -59,7 +61,7 @@ export const DosesDescriptionList = ({
       dosesPercent={dhbData.secondDosesPercentage}
     >
       <dd className={styles["vaccine-count"]}>
-        {numberFormatter.format(dhbData.secondDosesTo90Percent)} left
+        {numberFormatter.format(dhbData.secondDosesTo90Percent)} left{" "}
```
andycarrell, line 64: Rm space 
```suggestion
        {numberFormatter.format(dhbData.secondDosesTo90Percent)} left
```

andycarrell (approved):

app/components/DosesDescriptionList/DosesDescriptionList.tsx
andycarrell: Hehe and other one

finnclark (approved):

app/propsGenerators/indexPagePropsGenerator.ts
```diff
@@ -11,21 +11,31 @@ import {
   IndexPageProps,
 } from "../types/IndexPageProps";
 
+function calculateDosePercentage(
+  dosesCount: number,
+  totalPopulation: number
+): number {
+  return parseFloat(((dosesCount / totalPopulation) * 100).toFixed(6));
```
jishaal, line 18: `toFixed` still rounds, but 6dp should be "enough" precision for us to not have it cause an issue

PR #124, opened 2021-11-03
Our style 404 page

andycarrell:
_Note that this doesn't change the copy from the default, it adds our styling_

### Changes
- Adds styling, including light and dark modes
- Adds reduced footer content (todo: move footer to re-usable component)
- Adds a "return home" link (adds an internal link component to achieve this

### After
image

### Before
image

### DIFF
--- a/app/styles/typography.scss
+++ b/app/styles/typography.scss
@@ -34,7 +34,7 @@
   font-weight: 500;
 }
 
-@mixin footnote {
+@mixin p {
   font-size: 0.875rem;
   line-height: 1.25rem;
   font-weight: 400;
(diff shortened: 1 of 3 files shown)


app/styles/typography.scss
```diff
@@ -34,7 +34,7 @@
   font-weight: 500;
 }
 
-@mixin footnote {
+@mixin p {
```
andycarrell, line 37: Thoughts on this change? I wanted to create a class for "normal" text that we could easily use on pages (as opposed to components which define their own css).

It can definitely be updated (sizes, spacing etc) at a later date?

finnclark (approved):
Lgtm

PR #125, opened 2021-11-04
Send a notification whenever data is updated

theramis:
it's getting annoying checking github if data has been updated. 

This will send us a message on discord when the data is updated.

### DIFF
--- a/fetchy/fetchyboi.ts
+++ b/fetchy/fetchyboi.ts
@@ -2,40 +2,62 @@ import getVaccineData from "./vaccineDataService";
 import {
   createVaccineDataForDate,
   getVaccineDataForDate,
-  updateVaccineDataIfDifferentForDate,
+  updateVaccineDataForDate,
 } from "../shared/dataRepository";
 import got from "got";
+import { diff } from "deep-object-diff";
+
+const sendNotification = async (message: string) => {
+  // don't send notifications if running in a PR
+  if (process.env.GITHUB_EVENT_NAME === "pull_request") {
+    return;
+  }
+
+  await got.post(process.env.DISCORD_WEBHOOK!, {
+    json: {
+      avatar_url:
+        "https://i1.sndcdn.com/artworks-000497194077-gnjnca-t500x500.jpg",
+      content: message,
+      username: "Fetchy",
+    },
+  });
+};
+
+const sendDataUpdatedNotification = async () => {
+  await sendNotification(
+    ":raised_hands: :raised_hands: **Fetchy got some new data** :raised_hands: :raised_hands:"
+  );
+};
+
+const areObjectsDifferent = (a: object, b: object) =>
+  Object.keys(diff(a, b)).length !== 0;
 
 const main = async () => {
   try {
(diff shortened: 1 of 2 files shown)


andycarrell (approved):
Let's merge this in - will leave some comments that you can address now or later ;)

shared/dataRepository.ts
```diff
@@ -26,18 +26,11 @@ export const getVaccineDataForDate = async (
...
-    return;
-  }
-
-  if (areObjectsDifferent(existingData, data)) {
-    await createVaccineDataForDate(time, data);
-  }
+  await createVaccineDataForDate(time, data);
 };
 
 const areObjectsDifferent = (a: object, b: object) =>
```
andycarrell, line 36: Is this function still used in this file?

shared/dataRepository.ts
```diff
@@ -26,18 +26,11 @@ export const getVaccineDataForDate = async (
...
 ) => {
-  const existingData = await getVaccineDataForDate(time);
-  if (existingData == null) {
-    return;
-  }
-
-  if (areObjectsDifferent(existingData, data)) {
-    await createVaccineDataForDate(time, data);
-  }
+  await createVaccineDataForDate(time, data);
```
andycarrell, line 33: This is pass through, so, is there any difference between `createVaccineDataForDate` & `updateVaccineDataForDate`? If not can we just use `createVaccineDataForDate`?

fetchy/fetchyboi.ts
```diff
@@ -2,40 +2,62 @@ import getVaccineData from "./vaccineDataService";
...
+};
+
+const sendDataUpdatedNotification = async () => {
+  await sendNotification(
+    ":raised_hands: :raised_hands: **Fetchy got some new data** :raised_hands: :raised_hands:"
+  );
+};
+
+const areObjectsDifferent = (a: object, b: object) =>
+  Object.keys(diff(a, b)).length !== 0;
```
andycarrell, line 33:
```suggestion
  Boolean(Object.keys(diff(a, b)).length);
```

fetchy/fetchyboi.ts
```diff
@@ -2,40 +2,62 @@ import getVaccineData from "./vaccineDataService";
...
   getVaccineDataForDate,
-  updateVaccineDataIfDifferentForDate,
+  updateVaccineDataForDate,
 } from "../shared/dataRepository";
 import got from "got";
+import { diff } from "deep-object-diff";
+
+const sendNotification = async (message: string) => {
+  // don't send notifications if running in a PR
+  if (process.env.GITHUB_EVENT_NAME === "pull_request") {
```
andycarrell, line 12: Does this send notifications if you run `fetchyboi` locally? Maybe we should make this an opt in to messaging and then we set an env var in the github action?

```suggestion
  if (process.env.SEND_DISCORD_NOTIFICATIONS === "true") {
```

theramis: Ooo yes much better idea. Will do in another pr

PR #128, opened 2021-11-05
Update 2021-11-04.txt

jishaal:
Not sure if this will work but could be another solution until we derive the %


jishaal:
Prolly not great since the UI doesn't truncate 🤷🏼

PR #130, opened 2021-11-05
Start storing the raw html files of the MoH site

theramis:
### Changes 
- Changed the data files to have an extension of `json` rather than `txt` (this gives you nice syntax highlighting and is just more correct)
- Moved the data files from the `data` folder to the `data/vaccineData` folder. 
- Added support to store the rawHtml file as well to the `data/rawSite` folder

### Bug fixes
- The `sendNotification` method in fetchBoi was incorrectly going to send a notification when the send notification flag was set to FALSE. So fixed that

### Renames
- Renamed the `dataRepository` to `vaccineDataStore`. It also had some refactoring in there

### Refactors
- Renamed `getVaccineData` to `getAllVaccineData` since this will be used in the future to get more than just yesterday and the todays data.


shared/vaccineDataStore.ts
```diff
@@ -0,0 +1,54 @@
...
+  const files = await fs.readdir(CONSTANTS.vaccineDataFolder);
+
+  const vaccineData: VaccineData[] = [];
+  const filesToParse = files.sort().reverse().slice(0, maxItems);
+
+  for (const f of filesToParse) {
+    const filePath = `${CONSTANTS.vaccineDataFolder}/${f}`;
+    const data = await fs.readFile(filePath, "utf8");
+    vaccineData.push(JSON.parse(data));
+  }
```
andycarrell, line 39: Could use promise.all?

```suggestion
  const filesToParse = files.sort().reverse().slice(0, maxItems);
  const vaccineData: VaccineData[] = await Promise.all(filesToParse.map(async (f) => {
    const filePath = `${CONSTANTS.vaccineDataFolder}/${f}`;
    const data = await fs.readFile(filePath, "utf8");
    return JSON.parse(data);
  });
```

shared/vaccineDataStore.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return vaccineData;
```
andycarrell, line 41: Could use promise.all? Run in parallel?

```suggestion
  const files = await fs.readdir(CONSTANTS.vaccineDataFolder);
  const filesToParse = files.sort().reverse().slice(0, maxItems);

  const vaccineData = filesToParse.map(async (f) => {
    const filePath = `${CONSTANTS.vaccineDataFolder}/${f}`;
    const data = await fs.readFile(filePath, "utf8");
    return JSON.parse(data);
  });

  return Promise.all(vaccineData);
```

andycarrell (approved):

shared/vaccineDataStore.ts
theramis: ooo great idea

jishaal (approved):

PR #132, opened 2021-11-06
Feature - Adds last checked at, with fetchy's last run

andycarrell:
- Adds a "dataStore" for github. Not sure on naming here, open to suggestions
- Adds a route for /fetchy-boi/latest-run
- Deprecates /trigger-fetchy-boi for /fetchy-boi/trigger (@theramis - can you please update the CRON job if this PR lands?)

First pass at naming - looking for feedback here

### DIFF
--- a/app/types/IndexPageProps.ts
+++ b/app/types/IndexPageProps.ts
@@ -13,4 +13,5 @@ export type DhbVaccineDoseDataForIndexPage = DhbVaccineDoseData & {
 export type IndexPageProps = {
   allDhbsVaccineDoseData: DhbVaccineDoseDataForIndexPage[];
   dataValidAsAtTimeUtc: string;
+  lastCheckedAtTimeUtc: string | null;
 };
--- a/pages/api/fetchy-boi/latest-run.ts
+++ b/pages/api/fetchy-boi/latest-run.ts
@@ -1,0 +1,10 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+import { getLastestFetchyRun } from "../../../shared/githubDataStore";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  res.status(200).json({ data: await getLastestFetchyRun() });
+}
--- a/pages/api/fetchy-boi/trigger.ts
+++ b/pages/api/fetchy-boi/trigger.ts
@@ -1,0 +1,17 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+import { triggerFetchy } from "../../../shared/githubDataStore";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const secret = req.query["secret"];
+
+  if (secret !== process.env.API_SECRET) {
+    res.status(401).json({ message: "Missing or incorrect secret!" });
+  } else {
+    const response = await triggerFetchy();
+    res.status(response.status).end();
+  }
+}
--- a/shared/githubDataStore.ts
+++ b/shared/githubDataStore.ts
@@ -19,3 +19,17 @@ export const getLastestFetchyRun = async () => {
 
   return latestRun;
 };
+
+export const triggerFetchy = async () => {
+  const octokit = new Octokit({ auth: process.env.GITHUB_PAT });
+  const response = await octokit.request(
+    "POST /repos/{owner}/{repo}/dispatches",
+    {
+      owner: "theramis",
+      repo: "aucvid",
+      event_type: "trigger-fetchy-boi",
+    }
+  );
+
+  return response;
+};
--- a/shared/githubDataStore.ts
+++ b/shared/githubDataStore.ts
@@ -1,0 +1,21 @@
+import { Octokit } from "@octokit/core";
+
+export const getLastestFetchyRun = async () => {
+  const octokit = new Octokit({ auth: process.env.GITHUB_PAT });
+  const { data } = await octokit.request(
+    "GET /repos/{owner}/{repo}/actions/workflows/{workflow_id}/runs",
+    {
+      owner: "theramis",
+      repo: "aucvid",
+      workflow_id: "fetchy.yml",
+      branch: "main",
+      status: "success",
+      exclude_pull_requests: true,
+      per_page: 1,
+    }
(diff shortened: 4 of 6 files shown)


shared/githubDataStore.ts
```diff
@@ -0,0 +1,38 @@
+import { Octokit } from "@octokit/core";
+
+const githubOwner = "theramis";
+const githubRepo = "aucvid";
+
+const octokit = new Octokit({ auth: process.env.GITHUB_PAT });
```
andycarrell, line 6: All good to have this re-used?

app/types/IndexPageProps.ts
```diff
@@ -13,4 +13,5 @@ export type DhbVaccineDoseDataForIndexPage = DhbVaccineDoseData & {
 export type IndexPageProps = {
   allDhbsVaccineDoseData: DhbVaccineDoseDataForIndexPage[];
   dataValidAsAtTimeUtc: string;
+  lastCheckedAtTimeUtc: string | null;
```
theramis, line 16: Is there a scenario when it would be null?

pages/api/fetchy-boi/latest-run.ts
```diff
@@ -0,0 +1,10 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+import { getLastestFetchyRun } from "../../../shared/githubDataStore";
+
+export default async function handler(
```
theramis, line 5: Why do we need an API endpoint for this?
Are we going to get the UI call it as well as load it from the props with ssr?

pages/api/fetchy-boi/trigger.ts
```diff
@@ -0,0 +1,17 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+import { triggerFetchy } from "../../../shared/githubDataStore";
+
+export default async function handler(
```
theramis, line 5: I'm alg with the new route :)

shared/githubDataStore.ts
theramis: Hmm I don't have a great reason but seems "safer" to not use admin auth to make a call when auth is not needed.

andycarrell: Will double check I thought so too, but then saw some errors 🤔

app/types/IndexPageProps.ts
andycarrell: If we can't find a run? Am I being too defensive??

pages/api/fetchy-boi/latest-run.ts
andycarrell: Ah will need to double check with @jishaal here - is the plan to have the last checked at datetime update after the page has loaded? If so we need it. My thinking is this endpoint will drive the "new data" ux

jishaal: I think it was for the case where the page is left open / not refreshed and the "last checked" datetime gets stale, solution being polling this API to get latest and render.

In terms of the "new data" ux, would need another API/way to figure out when new data is available as this API wouldn't be sufficient

app/types/IndexPageProps.ts
andycarrell: _Actually_ just had a double check why I did this - the underlying type is `string | undefined` - so I'm unsure how to propagate that up to the app?

theramis: oops i just realised i was thinking of this as the last updated time not last checked. 

It's alg to have this as string | null/undefined then :)

shared/githubDataStore.ts
theramis: alg I'm not too fussed

app/types/IndexPageProps.ts
theramis: yea I'd say so. We can gaurentee it will always be there since we control the files. 
By having `null` the index page now needs to account for what to do when there is a null and there isn't a super obvious answer there too haha. 


But also I'm not fussed about this too much. So whatever you think is best go with it.

pages/api/fetchy-boi/latest-run.ts
theramis: ahh sweet

theramis (approved):

pages/api/fetchy-boi/latest-run.ts
andycarrell: Sweet. Sounds good to me.

I think we could _probably_ extend this API to have the data about last updated too?

app/types/IndexPageProps.ts
andycarrell: Yup I gotcha

My thinking was to not show the "last checked at" bit if the values aren't available - I think if it's `null` the whole page will fail (luxon will throw) which is less than ideal, especially since this is very much secondary data.

You're right about us having control - I'll update it to not be null, we can handle the case where it's not if it actually happens haha

jishaal: If we mark it as nullable i.e.  `lastCheckedAtTimeUtc?: string` then we don't need to have `null`  unless we want to be explicit?

pages/api/fetchy-boi/latest-run.ts
jishaal: Sounds good!

app/types/IndexPageProps.ts
andycarrell: I've spent 2 years being explicit 😅 - happy to go with the crowd though. Is `?` more common in typescript - that would make sense to me too 🤔 cc @theramis @jishaal

theramis: 🤷

jishaal: > ? more common in typescript - that would make sense to me too 

No hard or fast rule tbh, but generally is the pattern I follow. In this case you can make the call :)

jishaal: TS makes embracing `undefined` much easier 😉

jishaal (approved):

app/types/IndexPageProps.ts
andycarrell: Let's follow - will update in the next PR ❤️ 

Thanks team I'm on board with this

PR #133, opened 2021-11-07
Added older raw vaccine site data

theramis:
### Changes
- Used https://web.archive.org/web/*/https://www.health.govt.nz/our-work/diseases-and-conditions/covid-19-novel-coronavirus/covid-19-data-and-statistics/covid-19-vaccine-data to get historical versions of the MoH page. 
- I only went back to 22nd Oct as before that time the Vaccinations per DHB were not updated daily. 
- Also had to re-get 4th of November as it was added but then prettier had ran over it and it makes the regex break. 
- Added a prettier ignore so none of the new ones get formatted.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -19,3 +19,6 @@ yarn-error.log*
 
 # vercel
 .vercel
+
+# data html files
+data/rawSite/*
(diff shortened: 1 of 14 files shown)


jishaal:
Repo with all these files be like  😂 

image

jishaal (approved):

PR #134, opened 2021-11-07
Store created + updated date as part of VaccineData files (also includes ability to regen vaccine data files)

theramis:
## Changes
It's best to review this PR by commits IMO. Also probably turn on hide whitespace for differences. 

1. Created a concept of `DataDocument` which can store a generic object into the DataStore. It adds to the generic object some metadata (created and updated date)
2. Updated VaccineDataStore to use the `DataDocument` 
3. Unfortunately due to change #2 it meant reading the current json files doesn't work since the datastructure had changed. Rather than manually updating the files, I made fetchy have the ability to regenerate all of the files. This will be useful as we decide to extract more stuff from the page as we can regenerate our data structure from the raw site again. 

## TODO
- ~~Due to the regen, it deletes the data files which don't have rawSites at the moment. Wait until https://github.com/theramis/aucvid/pull/133 is merged and then regen the files for this PR.~~ **DONE**

### DIFF
--- a/app/propsGenerators/indexPagePropsGenerator.ts
+++ b/app/propsGenerators/indexPagePropsGenerator.ts
@@ -23,10 +23,12 @@ export default async function fetchIndexPageProps(): Promise<IndexPageProps> {
 
   return {
     allDhbsVaccineDoseData: generateAllDhbsVaccineDoseData(
-      latestData.vaccinationsPerDhb,
-      yesterdayData.vaccinationsPerDhb
+      latestData.data.vaccinationsPerDhb,
+      yesterdayData.data.vaccinationsPerDhb
     ),
-    dataValidAsAtTimeUtc: DateTime.fromISO(latestData.dataValidAsAtNzTimeIso)
+    dataValidAsAtTimeUtc: DateTime.fromISO(
+      latestData.data.dataValidAsAtNzTimeIso
+    )
       .toUTC()
       .toISO(),
   };
--- a/data/vaccineData/2021-10-29.json
+++ b/data/vaccineData/2021-10-29.json
@@ -1,215 +1,217 @@
 {
-  "vaccinationsPerDhb": [
-    {
-      "dhbName": "Northland",
-      "firstDoses": 127447,
-      "firstDosesPercentage": 0.79,
-      "firstDosesTo90Percent": 17741,
-      "secondDoses": 103000,
-      "secondDosesPercentage": 0.64,
-      "secondDosesTo90Percent": 42188,
-      "totalPopulation": 161320
-    },
-    {
-      "dhbName": "Waitemata",
-      "firstDoses": 478860,
-      "firstDosesPercentage": 0.91,
-      "firstDosesTo90Percent": 0,
-      "secondDoses": 417802,
-      "secondDosesPercentage": 0.79,
-      "secondDosesTo90Percent": 55676,
-      "totalPopulation": 526087
-    },
-    {
-      "dhbName": "Auckland",
-      "firstDoses": 397623,
-      "firstDosesPercentage": 0.94,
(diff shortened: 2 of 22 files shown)


fetchy/fetchyboi.ts
```diff
@@ -65,4 +82,12 @@ const main = async () => {
   }
 };
 
+const main = async () => {
+  if (process.env.REGEN_DATA === "true") {
```
andycarrell, line 86: Can we possibly be really explicit with this variable?

```suggestion
  if (process.env.REGENERATE_DATA === "true") {
```

fetchy/fetchyboi.ts
```diff
@@ -29,34 +31,49 @@ const sendDataUpdatedNotification = async () => {
   );
 };
 
-const areObjectsDifferent = (a: object, b: object) =>
-  Object.keys(diff(a, b)).length !== 0;
+const areObjectsDifferent = (a: Object, b: Object) =>
```
andycarrell, line 34: I think actually lowercase object is possibly better here - I sure don't know haha
```suggestion
const areObjectsDifferent = (a: object, b: object) =>
```

app/propsGenerators/indexPagePropsGenerator.ts
```diff
@@ -23,10 +23,12 @@ export default async function fetchIndexPageProps(): Promise<IndexPageProps> {
...
-      yesterdayData.vaccinationsPerDhb
+      latestData.data.vaccinationsPerDhb,
+      yesterdayData.data.vaccinationsPerDhb
     ),
-    dataValidAsAtTimeUtc: DateTime.fromISO(latestData.dataValidAsAtNzTimeIso)
+    dataValidAsAtTimeUtc: DateTime.fromISO(
+      latestData.data.dataValidAsAtNzTimeIso
+    )
       .toUTC()
       .toISO(),
```
andycarrell, line 33: Should this just be stored in UTC? Not part of this PR but something to consider 🙏

andycarrell:
PR looks good - if it were me, I'd create a completely separate function for regeneration - I don't really understand why we'd need to Regen based on config like that?

data/vaccineData/2021-10-29.json
```diff
@@ -1,215 +1,221 @@
...
-      "firstDosesPercentage": 0.12,
-      "firstDosesTo90Percent": 13416,
-      "secondDoses": 1653,
-      "secondDosesPercentage": 0.1,
-      "secondDosesTo90Percent": 13803,
-      "totalPopulation": 17173
-    }
-  ],
-  "dataValidAsAtNzTimeIso": "2021-10-29T23:59:00.000+13:00"
+  "metaData": {
```
jishaal, line 2: Minor but `metadata` should be lowercase

jishaal (commented):
LGTM. Cool to have the regen functionality now 👍🏼

jishaal (approved):

fetchy/fetchyboi.ts
theramis: sweet will do agreed.

fetchy/fetchyboi.ts
theramis: I have no idea haha, I think your commit made them upper case so I left it like that 😂

app/propsGenerators/indexPagePropsGenerator.ts
theramis: yea agree, will do some other time in a diff PR

theramis:
> PR looks good - if it were me, I'd create a completely separate function for regeneration - I don't really understand why we'd need to Regen based on config like that?

TBH same, I got lazy honestly

andycarrell (approved):

PR #145, opened 2021-11-08
Adds desktop layout (about 's' breakpoint) adds hidden utilities

andycarrell:


app/components/Page/Page.tsx
```diff
@@ -33,7 +33,7 @@ export const PageFooter = ({ className, children = null }: PageFooterProps) => (
     <div className="flex flex-row align-items-center justify-content-center footnote space-x-2xs">
       <div>&#128075;</div>
       <p>
-        Made <span hidden>with love</span> by{" "}
+        Made <span className="hidden s:visible">with love</span> by{" "}
```
andycarrell, line 36: Bringing back "with love"

app/styles/globals.scss
```diff
@@ -28,6 +28,22 @@ body {
   transition: background-color 0.3s ease-in-out;
 }
 
+// Hidden utilities
```
andycarrell, line 31: This gives us the ability to do `className="hidden s:visible"` of vice versa. That's pretty handy IMO - a couple of examples in this PR!

The downsides of this approach are:
- We only have access to "display: unset;" - this works for this use case but I don't think is a full solution
- This is a weird place to put hidden styling - it does _not_ play well with our `flex` className for example
- Hidden elements still render (have React lifecycles), they just don't appear in the dom

The alternative solution is to create a component that listens to the page width and applies a media query - the problem with this is it can only run on the client. Given we're server-side rendering, I think we need a CSS based solution too 🙏

pages/index.tsx
```diff
@@ -58,16 +58,21 @@ const Index: React.FC<IndexPageProps> = (props: IndexPageProps) => {
...
+              Data updated{" "}
+              {DateTime.fromISO(props.dataUpdatedAtTimeUtc).toRelative()}
             </p>
-          ) : null}
+            {props.lastCheckedAtTimeUtc ? (
+              <>
+                <span className="hidden s:visible margin-x-xs">-</span>
+                <p>
+                  Last checked{" "}
+                  {DateTime.fromISO(props.lastCheckedAtTimeUtc).toRelative()}
```
andycarrell, line 71: @jishaal - forgot to ask, does this mean we can delete https://github.com/theramis/aucvid/blob/main/app/utilities/hoursBeforeNow.ts?

jishaal: > forgot to ask, does this mean we can delete

I think so. Imo its note a huge issue rather get rid of the custom helper if we don't need it. We can use `toRelative({style: "short" })` if we want a shorter string (use `min` instead of `minutes`)

jishaal (approved):
Love it!

pages/index.tsx
andycarrell: Will do in another PR (alongside renaming the is / has mounted hook)

finnclark:
Sorry, yes looks great!

PR #148, opened 2021-11-09
RFC - Adds cache to github data store

andycarrell:
I'll be honest, I don't really know for sure if this is good, bad or good enough.

If someone has an easier implementation I'm here for 😅


shared/cache.ts
```diff
@@ -0,0 +1,47 @@
+import { DateTime } from "luxon";
+
+export type Cache<DataType> = {
```
andycarrell, line 3: Does this need to be exported?

shared/cache.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export const createCache = <DataType>(
+  defaultCache: Partial<Cache<DataType>> = {}
+) => {
+  let cache: Cache<DataType> = {
+    value: defaultCache.value ?? undefined,
+    updatedAt: defaultCache.updatedAt ?? undefined,
+    timeToLive: 5 * 60,
+  };
```
andycarrell, line 16: Creates a cache per usage (makes this re-usable)

shared/cache.ts
```diff
@@ -0,0 +1,47 @@
...
+          value,
+          updatedAt: DateTime.utc().toISO(),
+          timeToLive: cache.timeToLive,
+        };
+      } catch (e) {
+        console.error(`Error updating cache: ${e}`);
+      }
+    }
+
+    return cache.value;
```
andycarrell: Strategy is to return the cached value, even on failure - this is baked into the cache here. Not sure if we would want that in every case?

shared/githubDataStore.ts
```diff
@@ -1,25 +1,39 @@
...
 
 const githubOwner = "theramis";
 const githubRepo = "aucvid";
 
 const octokit = new Octokit({ auth: process.env.GITHUB_PAT });
 
+type LatestFetchyRunResponse = OctokitResponse<
+  { workflow_runs: { run_started_at?: string | undefined }[] },
+  200
+>;
```
andycarrell, line 14: Had to create this type to populate the cache - is this what I should / shouldn't be doing?

shared/githubDataStore.ts
```diff
@@ -1,25 +1,39 @@
...
+        exclude_pull_requests: true,
+        per_page: 1,
+      }
+    );
+
+    return response;
+  });
 
-  const [latestRun = null] = data.workflow_runs;
+  const [latestRun = null] = cachedValue?.data?.workflow_runs ?? [];
```
andycarrell: 😬 🙃 it is what it is :(

theramis:
TBH I understand very little of what is going on haha. 

Will checkout locally and see if I can make sense of it 😅

shared/cache.ts
theramis: nope don't see a need to export this.

theramis:
So based on how I understand JS this should work.

Thoughts on adding a log in there temporarily so we can see in prod when the cache is being hit vs when it's calling the real function?
Would be most realistic test to see if its working as expected in a lambda.

shared/githubDataStore.ts
```diff
@@ -1,25 +1,39 @@
 import { Octokit } from "@octokit/core";
+import { OctokitResponse } from "@octokit/types";
+
```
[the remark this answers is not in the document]
andycarrell: Yes, separating local & modules imports.

Sorry, old habit :(

andycarrell:
> I actually feel like caching should be added at the edge if possible.
So in this I'd say you should probably add the cache in index props generator.

@theramis - I agree. I'll look at opening a new PR I think 🙏

andycarrell:
Close in favour of #153 🙏

PR #149, opened 2021-11-09
Add endpoint to fetch latest metadata

jishaal:
### Changes

Added a new endpoint at `api/data/latest-metadata`. The idea is that we could use the endpoint to poll for changes from the client and show the "Updated Data Available, Please Refresh" message when new data is available. I don't think we should couple to the `fetchy-boi/latest-run` since we may want to avoid getting rate limited 😅 Might make sense to not have the API exposed/called client side?

We could poll from the client-side based on certain parameters, perhaps if they are idle for a long time? 

We could also be "smart" and use sockets or something else to push a notification to the client - is that worth exploring?


andycarrell (approved):
LGTM - to keep it a bit more generic / flexible, we could have an endpoint `/api/data/latest` that returns the file contents? 

Happy with PR as it though - decoupling from last updated at is probably a good idea

andycarrell:
> We could poll from the client-side based on certain parameters, perhaps if they are idle for a long time?
We could also be "smart" and use sockets or something else to push a notification to the client - is that worth exploring?

Regarding this, I think if we choose a modern data fetching library, we should get interval polling and tab focus refresh for free / with minimal config:
- SWR - https://swr.vercel.app/ (I swear I don't have any stock ;) )
- React Query - https://react-query.tanstack.com/

jishaal:
> LGTM - to keep it a bit more generic / flexible, we could have an endpoint /api/data/latest that returns the file contents?

Yeah was thinking that, will need to extract `fetchIndexPageProps` to have it go through the same mapping logic to not send "incorrect" data through. Reckon we need it?

> Regarding this, I think if we choose a modern data fetching library, we should get interval polling and tab focus refresh for free / with minimal config

Haha yes big fan of `react-query`. I mean in this case do we just 
- Create a `/api/data/latest` 
- Use `react-query` to fetch latest data on focus/at an interval?
- Skip the whole new "Updated Data Available, Please Refresh" UX? 

🤔

theramis:
If we are going to use this endpoint as the way to get the latest data. 
Then do you also wanna delete this endpoint as part of this PR?
https://github.com/theramis/aucvid/blob/main/pages/api/fetchy-boi/latest-run.ts

theramis (approved):

andycarrell:
Given our chat offline, I think it might be best to merge this and plough on?

I guess it depends how we're planning on "refreshing the data" when the user clicks "refresh"

jishaal:
> If we are going to use this endpoint as the way to get the latest data. Then do you also wanna delete this endpoint as part of this PR? https://github.com/theramis/aucvid/blob/main/pages/api/fetchy-boi/latest-run.ts

@andycarrell we delete this?

andycarrell:
@jishaal I'm unsure currently - does this endpoint replace it?

We can delete it though and add it back if we need haha

jishaal:
No it doesn't. I guess the idea was if we were going to poll for the latest run from the UI we were going to use that endpoint. But seeing we have some constraints in terms of rate limiting + moving to letting the user know when the data was stale, can remove it

andycarrell:
@jishaal yup correct let's 🔥🔥🔥 it

jishaal:
Nuked it

PR #150, opened 2021-11-09
Add All DHBs option to Region dropdown

jishaal:
### Changes

- Adds a new `All DHBs` option in the region selector to view all DHBs in a single list
- The "All DHBs" option does now expose the `Overseas/Unknown` group. Do we want to filter it out?


theramis (approved):

theramis:
> The "All DHBs" option does now expose the Overseas/Unknown group. Do we want to filter it out?
IMO we should keep it in for all

finnclark (approved):

PR #153, opened 2021-11-10
RFC - Cache (2.0) 

andycarrell:
Inspo from #148 - takes the feedback from there and iterates 🙏

### DIFF
--- a/app/propsGenerators/indexPagePropsGenerator.ts
+++ b/app/propsGenerators/indexPagePropsGenerator.ts
@@ -20,9 +20,9 @@ function calculateDosePercentage(
   return parseFloat(((dosesCount / totalPopulation) * 100).toFixed(6));
 }
 
-const getLatestFetchyRunCache = createCache(getLatestFetchyRun, {
+const getCachedLatestFetchyRun = createCache(getLatestFetchyRun, {
   key: "latest-fetchy-run",
-  ttl: 5 * 60,
+  ttlInSeconds: 5 * 60,
 });
 
 export default async function fetchIndexPageProps(): Promise<IndexPageProps> {
@@ -30,7 +30,7 @@ export default async function fetchIndexPageProps(): Promise<IndexPageProps> {
     { data: latestData, metadata: latestMetaData },
     { data: yesterdayData },
   ] = await getAllVaccineData(2);
-  const { value: latestFetchyRun } = await getLatestFetchyRunCache();
+  const { value: latestFetchyRun } = await getCachedLatestFetchyRun();
 
   return {
     lastCheckedAtTimeUtc: latestFetchyRun?.run_started_at ?? null,
--- a/app/propsGenerators/indexPagePropsGenerator.ts
+++ b/app/propsGenerators/indexPagePropsGenerator.ts
@@ -11,6 +11,7 @@ import {
   DhbVaccineDoseDataForIndexPage,
   IndexPageProps,
 } from "../types/IndexPageProps";
+import { createCache } from "../../shared/cache";
 
 function calculateDosePercentage(
   dosesCount: number,
@@ -19,16 +20,20 @@ function calculateDosePercentage(
   return parseFloat(((dosesCount / totalPopulation) * 100).toFixed(6));
 }
 
+const getLatestFetchyRunCache = createCache(getLatestFetchyRun, {
+  name: "latest-fetchy-run",
+  timeToLive: 5 * 60,
+});
+
 export default async function fetchIndexPageProps(): Promise<IndexPageProps> {
   const [
     { data: latestData, metadata: latestMetaData },
     { data: yesterdayData },
   ] = await getAllVaccineData(2);
-  const latestFetchyRun = await getLatestFetchyRun();
-  const { run_started_at: lastCheckedAtTimeUtc } = latestFetchyRun ?? {};
+  const { value: latestFetchyRun } = await getLatestFetchyRunCache();
 
   return {
-    lastCheckedAtTimeUtc,
+    lastCheckedAtTimeUtc: latestFetchyRun?.run_started_at ?? null,
     dataUpdatedAtTimeUtc: latestMetaData.updatedAtUtcTimeIso,
     allDhbsVaccineDoseData: generateAllDhbsVaccineDoseData(
       latestData.vaccinationsPerDhb,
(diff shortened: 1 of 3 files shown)


app/types/IndexPageProps.ts
```diff
@@ -14,5 +14,5 @@ export type IndexPageProps = {
   allDhbsVaccineDoseData: DhbVaccineDoseDataForIndexPage[];
   dataValidAsAtTimeUtc: string;
   dataUpdatedAtTimeUtc: string;
-  lastCheckedAtTimeUtc?: string;
+  lastCheckedAtTimeUtc: string | null;
```
andycarrell, line 17: This has to be `null` - because it's serialized, so it can't be `undefined`. Decided to be explicit here, fixes it!

app/propsGenerators/indexPagePropsGenerator.ts
```diff
@@ -19,16 +20,20 @@ function calculateDosePercentage(
   return parseFloat(((dosesCount / totalPopulation) * 100).toFixed(6));
 }
 
+const getLatestFetchyRunCache = createCache(getLatestFetchyRun, {
+  name: "latest-fetchy-run",
+  timeToLive: 5 * 60,
```
jishaal, line 25: Minor/preference: shall we just use the abbreviation here? Feel like `ttl` is fairly common?

shared/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const defaultTimeToLive = 5 * 60;
+
+export const createCache = <DataType>(
+  getValue: () => Promise<DataType>,
+  defaultCache: Partial<Cache<DataType>> = {}
+) => {
+  let cache: Cache<DataType> = {
+    value: defaultCache.value ?? undefined,
+    updatedAt: defaultCache.updatedAt ?? undefined,
```
jishaal, line 18: Would we want to provide this from the consumer?

shared/cache.ts
```diff
@@ -0,0 +1,70 @@
+import { DateTime } from "luxon";
+
+type Cache<DataType> = {
+  value?: DataType;
+  updatedAt?: string;
+  timeToLive: number;
+  name?: string;
```
jishaal, line 7: Thoughts on `key` instead of `name`?

shared/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      try {
+        const value = await getValue();
+
+        cache.value = value;
+        cache.updatedAt = DateTime.utc().toISO();
+
+        console.log(`${cache.name ? `${cache.name} - ` : ""}Cache updated.`);
+        return cache;
+      } catch (e: any) {
```
jishaal, line 37: Hehe `any`

jishaal (approved):
Minor comments, apart from that this is very cool! 😲

shared/cache.ts
andycarrell: Hehe is there any (it you will) better way to do it?

andycarrell:
Thanks @jishaal afk atm but will address when I get home 🙏

shared/cache.ts
andycarrell: I think if we're pre-populating the cache (ie default value) we'd want to set a default 'updatedAt', so we can set when the cache will expire 🙏 Maybe 'updatedAt' doesn't accurately describe it?

Otherwise, we could set the "default" to be more of a "config" with just `key` and `ttl`

shared/cache.ts
andycarrell: https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript I suppose ;)

shared/cache.ts
jishaal: `populatedAt` which you have renamed to works, maybe it could be `populatedAtUtc` if we we want to be more specific?

shared/cache.ts
jishaal: Haha imo leave as is if ya want

andycarrell: Might look at that in a future PR. Seems like it'd be good to have / know 🤩

jishaal:
Merge away once happy 👍🏼

app/propsGenerators/indexPagePropsGenerator.ts
```diff
@@ -19,16 +20,20 @@ function calculateDosePercentage(
   return parseFloat(((dosesCount / totalPopulation) * 100).toFixed(6));
 }
 
+const getLatestFetchyRunCache = createCache(getLatestFetchyRun, {
```
theramis, line 23: thoughts on `getLatestFetchyRunCache` -> `getCachedLatestFetchyRun`

shared/cache.ts
```diff
@@ -0,0 +1,61 @@
+import { DateTime } from "luxon";
+
+type Cache<DataType> = {
+  value?: DataType;
+  populatedAt?: string;
+  ttl: number;
```
theramis, line 6: thoughts on saying `ttlInSeconds` so peeps know this is in seconds not mins or hours. 

Not too fussed tbh so feel free to ignore.

shared/cache.ts
```diff
@@ -0,0 +1,61 @@
+import { DateTime } from "luxon";
+
+type Cache<DataType> = {
+  value?: DataType;
+  populatedAt?: string;
```
theramis, line 5: Not married to this but thoughts on making this a DateTime type?
I don't see the benefit of storing it as a string and then having to constantly convert it to a datetime object.

shared/cache.ts
andycarrell: `timeToLiveInSeconds` 😅 

Happy to update ;)

shared/cache.ts
andycarrell: I think it's best if we store it as a string. There's a couple of benefits to that (although typescript might make them unnecessary?):
- It's clear that _every_ datetime property needs to be converted first, so it's consistent everywhere
- We know we can serialize it, if we ever need to send it across the wire (note that if it every makes it to page props it'll need to be `json` serializable

Neither of those advantages _really_ apply in this case, but I think we should follow a strategy that's consistent everywhere, so we're not guessing why a particular part is different.

I also don't think there's too much overhead in converting it - since anywhere we convert it we need to use other luxon utilities (like formatting or getting time differential) anyway.

theramis: yea lets do the audit in a different PR. 

I agree with some of you above points but not really fussed so happy to go with it :)

theramis (approved):

theramis:
image

Caching is working with the deploy preview. 
Lets leave the console logs in just in case. No harm right?

andycarrell:
Yeh I think so @theramis - not like there's many eyes on the logs at this point anyway

PR #155, opened 2021-11-10
Make fetchy alert if can't parse the data correctly

theramis:
So this afternoon when fetchy was constantly getting new data. It was trying to parse a number and failing. Then when it tried to compare a `NaN` to a `null` it finds a difference and repeats. 

This all happened because the MoH site added these two lines at the bottom of the table. 
image

The "fix" in this PR will make it so that if it can't parse the content from the table into a number it will crash and alert us. 
This fixes us from making sure that any new rows or if they change anything in the table and we can't parse it, then it would alert us. 

1. This approach gives us some flexibility like if they add more Dhbs (highly unlikely) that it would keep working but it more prone to randomly introducing new data/breaking.

2. Another approach we can go down is to code in EXACTLY what we expect. If we get something other than that, then alert. 
e.g. make sure we get only the dhbs we have in code, if we get anything else we crash. etc etc. 

Thoughts? 

Do we want the more strict approach (i.e. number 2) or okay with 1 for now?

### DIFF
--- a/fetchy/utilities.ts
+++ b/fetchy/utilities.ts
@@ -1,2 +1,8 @@
-export const convertToNumber = (s: string): number =>
-  Number(s.replace(/[%,]/g, ""));
+export const convertToNumber = (s: string): number => {
+  const stripped = s.replace(/[%,]/g, "");
+  const result = Number(stripped || NaN);
+  if (Number.isNaN(result)) {
+    throw new Error(`Could not convert string ${s} to a number.`);
+  }
+  return result;
+};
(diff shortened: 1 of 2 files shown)


andycarrell:
> Do we want the more strict approach (i.e. number 2) or okay with 1 for now? 

I think number 1 is fine - we might get burnt again I suppose, if we do we can re-assess haha

fetchy/utilities.ts
```diff
@@ -1,2 +1,8 @@
-export const convertToNumber = (s: string): number =>
-  Number(s.replace(/[%,]/g, ""));
+export const convertToNumber = (s: string): number => {
+  const result = Number(s.replace(/[%,]/g, ""));
+  console.log(result);
+  if (isNaN(result) || result === null || result === undefined) {
+    throw new Error(`Could not convert string ${s} to a number.`);
+  }
+  return result;
```
andycarrell, line 7: I don't think `Number(x)` can return `null` or `undefined` so those checks are probably redundant. Also Number("") returns `0` (is that what we want??)

Soooo I think we might want something more like:

```suggestion
  const stripped = s.replace(/[%,]/g, ""); // strip known non-digit characters
  const result = Number(stripped || NaN); // convert to number or NaN by default
  if (Number.isNaN(result)) {
    throw new Error(`Could not convert string ${s} to a number.`);
  }
  return result;
```

jishaal (approved):

PR #156, opened 2021-11-13
If greater than 99, show > 99%

andycarrell:


app/components/DosesDescriptionList/DosesDescriptionList.tsx
```diff
@@ -16,6 +18,14 @@ type DosesDescriptionProps = {
 
 type DosesDescriptionListProps = { dhbData: DhbVaccineDoseDataForIndexPage };
 
+const DisplayPercentage = ({ percentage }: DisplayPercentageProps) => {
```
jishaal, line 21: Imo its fine to inline these especially for these smaller components e.g. 

const DisplayPercentage = ({ percentage }: { percentage: number } )

jishaal (approved):

theramis (approved):

PR #157, opened 2021-11-13
Store raw percentage data as it is not always a number

theramis:
### Changes 
There are a few small changes in this PR. 

1. Rather than taking the percentage from the table and storing it as a decimal. Now storing the raw string. We are not actually using this value so it didn't break anything. This avoids the problem of when the percentage is `>90%` which we can't easily  convert to a number. 
2. Renamed `firstDosesPercentage` to `firstDosesPercentageRaw` and added new props to replace `firstDosesPercentage` on the `DhbVaccineDoseDataForIndexPage` type`
3. Since first change changed how we read/store data had to regen all data. Hence all the changes in files. 
4. Log Fetchy notifications everytime rather than it is not sending notifications to discord. This helps us see what notification was sent as part of the run in github. 
5. Log the differences in console. This would help us see the diff in github logs. 

Not sure if this is the best approach to fix the bug we have with `>90%` but seems legit. 
The other option is to remove the raw percentage from our data completely since we don't use it. Can always add it back when/if we need it


data/vaccineData/2021-10-26.json
```diff
@@ -1,221 +1,221 @@
...
-        "secondDosesPercentage": 0.09,
+        "secondDosesPercentageRaw": "9%",
         "secondDosesTo90Percent": 13850,
         "totalPopulation": 17173
       }
     ],
     "dataValidAsAtNzTimeIso": "2021-10-26T23:59:00.000+13:00"
   }
-}
+}
```
andycarrell, line 221: 👍

data/vaccineData/2021-10-30.json
```diff
@@ -1,221 +1,221 @@
 {
   "metadata": {
-    "createdAtUtcTimeIso": "2021-11-07T05:46:24.787Z",
-    "updatedAtUtcTimeIso": "2021-11-07T05:46:24.787Z"
+    "createdAtUtcTimeIso": "2021-11-13T08:14:28.980Z",
+    "updatedAtUtcTimeIso": "2021-11-13T08:14:28.980Z"
```
andycarrell, line 4: :( Should these have changed?

andycarrell (approved):

data/vaccineData/2021-10-30.json
theramis: yea the files get completely regenerated. It deletes all the stored files and then recreates them. 

I could make the script only update the data if its different if ya prefer?

We didn't really use the created at and updated at date for much other than the last updated date on the site soo I didn't right a great regen script. Can do though if we want.

andycarrell: Yeh what do you think? I mean technically this is a re-intepretation of the same old data. Maybe the updated / created time should live with the raw html?

It may be an issue if we try and create pages for previous dates 🤔

theramis: Yea I agree it should live with the html. Will have a look moving the date there somehow. 

Although I doubt we'd ever use the date for previous day's. Unless we wanna tell the user what time we fetched the data on. Which seems weird.

PR #158, opened 2021-11-13
Adds 'useQueryParam' wrapper on useState

andycarrell:
This is a simple, fairly restrictive version of query param handling for the index page. The approach is to store the initial value for a given param key in React state, and sync the query param in a useEffect when the state updates. To keep the path name as clean as possible, if the param value isn't valid, or if it matches the default value, it will be stripped from the url.

_Note: I tried to use query params directly, instead of syncing with state in between, but it re-rendered the whole page, and was a bit janky._

Because query params can be literally any string (or as Next parses them `string | string[] | undefined`), we need a way to ensure the parsed value is `RegionOptionId`. I've created a function in `RegionDropdown` to check the type, and return a default value if the type doesn't match. 

Question: Is this type checking or default value overkill? The other option is to assert the type is `RegionOptionId` and basically accept it will be possible wrong at runtime?

This approach uses Nextjs `useRouter`, which is nice because it encodes & decodes query params by default, but at the same time, it's pretty punishing since:
- The query object includes pathname variables (ie for `/blog/{blogId}/` - blogId would be in query object)
- There's no simple way of merging query params
- The push function isn't correctly memoized 👀 

Because of these shortcomings, and the fact that this hook only handles a single param, I think we need to update the implementation - which can be done in a later PR. I think it will be worth using `URLSearchParams` and handling some of the parsing & merging ourselves 🙏 

### Improvements
- Update hook to be "useQueryParams" to deal with the whole object of params
- Investigate type-safe query param parsing - is there a library to do this for us?

### Previously discussed
- ~Would it be better as `useQueryParams` that did the whole object?~ - Yes
- ~This method means there's always a query param - would it be better if we could have 'naked' paths, with no query params?~ - Ideally 'naked' routes. This PR has been updated
- ~peep the naming - I think "dhbs" makes the most sense?~ - confirmed!

### DIFF
--- a/app/hooks/useQueryParam.ts
+++ b/app/hooks/useQueryParam.ts
@@ -7,14 +7,14 @@ import {
 } from "react";
 import { useRouter } from "next/router";
 
-const useEventCallback = (fn: (...a: any[]) => any) => {
+const useEventCallback = <T extends (...args: any[]) => any>(fn: T): T => {
   const fnRef = useRef(fn);
 
   useLayoutEffect(() => {
     fnRef.current = fn;
   });
 
-  return useCallback((...args) => fnRef.current(...args), []);
+  return useCallback((...args: any[]) => fnRef.current(...args), []) as T;
 };
 
 export const useQueryParam = <ParamType extends string>(
--- a/app/hooks/useQueryParam.ts
+++ b/app/hooks/useQueryParam.ts
@@ -1,0 +1,46 @@
+import {
+  useState,
+  useEffect,
+  useRef,
+  useLayoutEffect,
+  useCallback,
+} from "react";
+import { useRouter } from "next/router";
+
+const useEventCallback = (fn: (...a: any[]) => any) => {
+  const fnRef = useRef(fn);
+
+  useLayoutEffect(() => {
+    fnRef.current = fn;
+  });
+
+  return useCallback((...args) => fnRef.current(...args), []);
+};
+
+export const useQueryParam = <ParamType extends string>(
+  key: string,
+  defaultValue: ParamType,
+  checkType: (s: unknown) => s is ParamType
+): [ParamType, React.Dispatch<React.SetStateAction<ParamType | undefined>>] => {
+  const router = useRouter();
+  const push = useEventCallback(router.push);
+
+  const [value, setValue] = useState<ParamType | undefined>(() => {
+    const initialQueryValue = router.query[key];
+    if (checkType(initialQueryValue)) {
+      return initialQueryValue as ParamType;
+    }
+    return undefined;
+  });
+
+  useEffect(() => {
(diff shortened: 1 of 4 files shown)


pages/index.tsx
```diff
@@ -13,16 +14,42 @@ import {
 } from "../app/components/Page";
 import {
   RegionDropdown,
+  isValidRegionId,
```
andycarrell, line 17:
```suggestion
```

jishaal:
> Would it be better as useQueryParams that did the whole object?

Yeah I think this would make sense since in the case there were multiple params. Not sure if we need it yet?

> This method means there's always a query param - would it be better if we could have 'naked' paths, with no query params?

Since the default is Auckland, I think this would be nicer in terms of a sharing perspective. So anything different than the default pushes a param

> peep the naming - I think "dhbs" makes the most sense?

I'm good with `dhbs`

andycarrell:
> Yeah I think this would make sense since in the case there were multiple params. Not sure if we need it yet?

We'll need it as soon as we have multiple params (this implementation replaces all query params unfortunately). The whole thing has been a bit of a minefield to be honest (especially with typescript layered in :( )

I think we could probably merge a first implementation (I've got a couple of other things to tidy up) and then iterate with a decent params implementation

jishaal:
> I think we could probably merge a first implementation 

Yep lets do it once you're happy 👍🏼

app/hooks/useQueryParam.ts
```diff
@@ -0,0 +1,46 @@
+import {
+  useState,
+  useEffect,
+  useRef,
+  useLayoutEffect,
+  useCallback,
+} from "react";
```
andycarrell, line 7: Going for the full set ;)

@jishaal - do you prefer `React.use...` or the separate imports here?

app/hooks/useQueryParam.ts
```diff
@@ -0,0 +1,46 @@
...
+  return useCallback((...args: any[]) => fnRef.current(...args), []) as T;
+};
+
+export const useQueryParam = <ParamType extends string>(
+  key: string,
+  defaultValue: ParamType,
+  checkType: (s: unknown) => s is ParamType
+): [ParamType, React.Dispatch<React.SetStateAction<ParamType | undefined>>] => {
+  const router = useRouter();
+  const push = useEventCallback(router.push);
```
andycarrell, line 26: Note: https://github.com/vercel/next.js/issues/18127

app/hooks/useQueryParam.ts
jishaal: Hehe, I prefer this way personally though no strong opinion on either. Just less words haha

andycarrell:
Prefer #165

PR #159, opened 2021-11-13
Added new relic browser

theramis:
Just wanted to give this a go and see what we get. 

No other reason than curiosity tbh. 
Happy to not have it if its not a good idea 😅


app/newRelicSnippet.tsx
```diff
@@ -0,0 +1,16 @@
...
+;NREUM.loader_config={accountID:"3340026",trustKey:"3340026",agentID:"1385887337",licenseKey:"NRJS-94c48cdf4bf4fa5718d",applicationID:"1385887337"}
+;NREUM.info={beacon:"bam.nr-data.net",errorBeacon:"bam.nr-data.net",licenseKey:"NRJS-94c48cdf4bf4fa5718d",applicationID:"1385887337",sa:1}
+`;
+
+export const NewRelicSnippet = () => {
+  return (
+    <script
```
andycarrell, line 11: Thoughts on using next/Script here? https://nextjs.org/docs/basic-features/script

andycarrell:
What's it give us? Analytics and error reporting?

theramis:
> What's it give us? Analytics and error reporting?

Quite a few things. 
errors, performance, everything GA gives us  

image

jishaal:
Its cool! Not sure if we need it? Don't mind either way haha if we just want to use it as a means to explore

app/newRelicSnippet.tsx
```diff
@@ -0,0 +1,18 @@
...
+;NREUM.loader_config={accountID:"3340026",trustKey:"3340026",agentID:"1385887337",licenseKey:"NRJS-94c48cdf4bf4fa5718d",applicationID:"1385887337"}
+;NREUM.info={beacon:"bam.nr-data.net",errorBeacon:"bam.nr-data.net",licenseKey:"NRJS-94c48cdf4bf4fa5718d",applicationID:"1385887337",sa:1}
```
jishaal, line 8: Would we want keys between local/deploy-preview and prod to be different?

theramis: yea I wasn't sure about that. 
tbh I don't think it matters too much

theramis:
> Its cool! Not sure if we need it? Don't mind either way haha if we just want to use it as a means to explore

tbh I don't reallllly need it. 
Was just looking at new relic pricing and found out it was free so thought why not haha

app/newRelicSnippet.tsx
andycarrell: IMO we should. Can be as simple as not rendering the script locally?

app/newRelicSnippet.tsx
theramis: So I tried by just converting this to Script. (I might be doing it wrong)
I can't seem to get it to load on the page though. 

I tried even just adding this 

image


Added that to the _document, _app, index.tsx. 

Nothing. What am I doing wrong?

app/newRelicSnippet.tsx
theramis: Whats the best way to do that?

Is there an inbuilt way or should I use env variables?

andycarrell: @theramis just use Env variables. Probably can check if it's production at this point

app/newRelicSnippet.tsx
theramis: nvm figured it out, I was testing it incorrectly 😓

app/newRelicSnippet.tsx
theramis: done

app/newRelicSnippet.tsx
```diff
@@ -0,0 +1,21 @@
...
+  if (process.env.NEXT_PUBLIC_ENVIRONMENT === "production") {
+    return (
+      <Script
+        id="new-relic-snippet"
+        dangerouslySetInnerHTML={{ __html: snippet }}
+      />
+    );
+  }
+  return null;
+};
```
andycarrell, line 21: Damn this file name should be `NewRelicSnippet`

andycarrell (approved):

PR #160, opened 2021-11-14
Update data refresh button

jishaal:
#### Changes
- Adds refresh cta button that appears when updated data is available
- Uses `react-query` to query and store the `latest-metadata`, mainly because I'm more familiar with it than `swr`. It invalidates the cache after 5mins so it doesn't keep polling on refocus etc ([see react-query defaults](https://react-query.tanstack.com/guides/important-defaults)) 

**DRAFT PR**
- Mainly because a few things on colors etc, especially hover, light theme for both the button and the mask. Maybe animation too. Need your help on that one @finnclark 🙏🏼 
- The check for the button to appear is commented out so you folks can actually see it appear in the preview 😅


app/components/LatestDataRefresh/LatestDataRefresh.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const getLatestMetadata = async (): Promise<LatestMetadataResponse> =>
+  clientFetcher("/api/data/latest-metadata");
+
+export const LatestDataRefresh = ({
+  className,
+  dataUpdatedAtTimeUtc,
+}: LatestDataRefreshProps) => {
+  const { data, error } = useQuery("latest-metadata", getLatestMetadata, {
```
jishaal, line 21: Should probably do something with [`initialData` config](https://react-query.tanstack.com/guides/initial-query-data#using-initialdata-to-prepopulate-a-query) here so it doesn't do a call on initial mount/render

jishaal: I also went with the more "simpler" approach here https://react-query.tanstack.com/guides/ssr#using-nextjs

andycarrell: I think this approach is fine 🙌 No real need to over-optimise it right now. 

Also, I think we'd maybe want to refactor the metadata endpoint (so we can re-use the same logic). Probably best to do that in a separate PR

app/utilities/clientFetcher.ts
```diff
@@ -0,0 +1,12 @@
+import { ErrorResponse } from "../../shared/types/api";
+
+const clientFetcher = (url: string) =>
+  fetch(url).then(async (response) => {
+    if (!response.ok) {
+      const errorMessage = (await response.json()) as ErrorResponse;
+      throw new Error(errorMessage.message);
+    }
+    return response.json();
+  });
```
andycarrell, line 10: Any reason not to use async await here?

```suggestion
const clientFetcher = async (url: string) => {
  const response = await fetch(url);
  const data = await response.json();
  
  if (response.ok) {
    return data;
  } else {
    throw new Error((data as ErrorResponse).message);
  }
};
```

pages/_app.tsx
```diff
@@ -33,7 +37,10 @@ function MyApp({ Component, pageProps }: AppProps) {
         <meta property="twitter:description" content={metaTags.description} />
         <meta property="twitter:image" content={metaTags.imageUrl} />
       </Head>
-      <Component {...pageProps} />
+      <QueryClientProvider client={queryClient}>
+        <Component {...pageProps} />
+        <ReactQueryDevtools initialIsOpen={false} />
```
andycarrell, line 42: Do we leave this in in the production build?

andycarrell:
> Uses react-query to query and store the latest-metadata

I'm a big fan too! Sounds good to me :D

app/components/Button/Button.module.scss
```diff
@@ -0,0 +1,13 @@
+@import "../../styles/vars.scss";
+
+.button {
+  background: var(--button-background);
+  color: var(--button-foreground);
+  border-radius: $radius-full;
+  font-weight: 500;
+
```
andycarrell, line 8: Thoughts on a transition here?

```suggestion
  transition: color,background-color .2s ease;

```

finnclark: +1 on the same transition as the icon button for dark mode toggle :)

app/components/Button/Button.module.scss
```diff
@@ -0,0 +1,13 @@
+@import "../../styles/vars.scss";
+
+.button {
+  background: var(--button-background);
```
finnclark, line 4: Thinking we could use the same css vars as the dark mode toggle for this button actually. Might just be worth renaming it to something more generic?

But if you wanted to keep the seperate css vars, can you please follow the same bg/hover colours as the toggle? I know it's different from the design, sorry.

This one -> `--dark-mode-toggle`

And same for the hover -> `--dark-mode-toggle-hover`

app/components/LatestDataRefresh/LatestDataRefresh.module.scss
```diff
@@ -0,0 +1,23 @@
...
+  position: sticky;
+  width: 100%;
+  top: 0;
+  z-index: 1;
+
+  &:before {
+    content: "";
+    top: 0;
+    position: absolute;
+    height: 6.25rem;
```
finnclark, line 13:
```suggestion
    height: 6rem;
```

app/components/LatestDataRefresh/LatestDataRefresh.module.scss
```diff
@@ -0,0 +1,23 @@
...
+    width: 100%;
+    background: linear-gradient(180deg, #000000 0%, rgba(0, 0, 0, 0) 100%);
+    pointer-events: none;
+  }
+}
+
+.refresh-button {
+  position: absolute;
+  top: 3.75rem;
```
finnclark, line 22:
```suggestion
  top: 2rem;
```

app/components/LatestDataRefresh/LatestDataRefresh.module.scss
```diff
@@ -0,0 +1,23 @@
...
+  top: 0;
+  z-index: 1;
+
+  &:before {
+    content: "";
+    top: 0;
+    position: absolute;
+    height: 6.25rem;
+    width: 100%;
+    background: linear-gradient(180deg, #000000 0%, rgba(0, 0, 0, 0) 100%);
```
finnclark, line 15: Could this change to a white gradient on light mode please?
Might be needing another css var

finnclark (dismissed):
Looks great! Just a couple tweaks if that's ok!

app/components/Button/Button.module.scss
andycarrell: Ah should've known there would already be a transition in there 😏

pages/_app.tsx
jishaal: It shouldn't show in production mode by default :D 

> By default, React Query Devtools are only included in bundles when process.env.NODE_ENV === 'development', so you don't need to worry about excluding them during a production build.

https://react-query.tanstack.com/devtools

app/components/LatestDataRefresh/LatestDataRefresh.module.scss
jishaal: Yeah I am a little unsure on the light mode gradient, have a look and suggest ta

app/utilities/clientFetcher.ts
jishaal: Good point, will update

finnclark:
Ohhh sorry I wasn't super clear my bad. Ideally the bg color of the button is the same by color of the icon (so white button with black text for dark mode, black button with white text for light mode)

finnclark:
Same as what you had before but just using the same styles as the icon color/hover

jishaal:
Ah! And is it this on hover? 
image

Sorry still not clear on the hover styles 😅

app/components/LatestDataRefresh/LatestDataRefresh.tsx
```diff
@@ -0,0 +1,50 @@
...
+    staleTime: 5 * 60 * 1000,
+  });
+  const router = useRouter();
+
+  if (!data || error) {
+    return null;
+  }
+
+  // if (dataUpdatedAtTimeUtc === data.updatedAtUtcTimeIso) {
```
jishaal, line 30: Once its approved, this needs to be uncommented before merging

jishaal:
@finnclark lemme know, if there are still changes I can let you tweak them post merge, might be more efficient 😂

finnclark (approved):

PR #165, opened 2021-11-16
Adds 'useQueryParams' - so we can store state in url params

andycarrell:
Following learnings from #158, and inspired by https://github.com/amannn/next-query-params, this PR adds:
- `use-query-params` &
- `query-string`

To handle parsing, state management and interactions with query strings. 

---

This solves a number of problems with #158:
- Adds support for multiple query params all at once
- Simplifies the type assertion method

---

To achieve this approach, we need to:
- Add a NextQueryParamProvider - as an adaptor to `use-query-params` QueryParamProvider, passing in push and replace methods, as well as parsing the query (search) from the router object
- Update the region dropdown to properly reset the selected value on mount if it's not correct. @jishaal - this handles invalid attributes passed down. Let me know what you think, it's quite defensive but prevents an inconsistent UI


app/hooks/useFreeze.ts
```diff
@@ -0,0 +1,13 @@
...
+  initializer: FrozenType | (() => FrozenType)
+): FrozenType => {
+  const [frozenValue] = useState<FrozenType>(initializer);
+
+  return frozenValue;
+};
+
+export const useFreezeFunction = <FnType extends (...args: any[]) => any>(
+  fn: FnType
+) => useFreeze(() => fn);
```
andycarrell, line 13: A couple of hooks added in I commonly use (although never in ts lol). Probably overkill but nice to have IMO

pages/index.tsx
```diff
@@ -37,20 +37,20 @@ const Index: React.FC<IndexPageProps> = (props: IndexPageProps) => {
...
             <DhbsVaccineDoseDataList
-              key={region}
+              key={query.dhbs}
               dhbsVaccineDoseData={allDhbsVaccineDoseData.filter((dhb) => {
-                if (region === "all") {
+                if (query.dhbs === "all") {
                   return true;
                 }
-                return dhb.regionIds.includes(region);
+                return dhb.regionIds.includes(query.dhbs as DhbRegionId);
```
andycarrell, line 53: Need to cast since dhbs can be a string - should be fine since `.includes` won't fail 🙏

pages/index.tsx
```diff
@@ -11,18 +11,18 @@ import {
...
 import { DhbsVaccineDoseDataList } from "../app/components/DhbsVaccineDoseDataList";
 
 const Index: React.FC<IndexPageProps> = (props: IndexPageProps) => {
   const { allDhbsVaccineDoseData } = props;
+
   const hasMounted = useHasMounted();
-  const [region, selectedRegion] = useState<RegionOptionId>("auckland");
+  const [query, setQuery] = useQueryParams({
+    dhbs: withDefault(StringParam, "auckland"),
+  });
```
andycarrell, line 25: Pre-emptively `useQueryParams` in prep for search & filter ;)

app/components/RegionDropdown/RegionDropdown.tsx
```diff
@@ -16,14 +17,24 @@ const options: Option[] = [
...
 }: RegionDropdownProps) => {
+  useEffectOnce(() => {
+    const isValidSelectedRegion =
+      selectedRegion &&
+      options.map((o) => o.id as string).includes(selectedRegion);
+
+    if (!isValidSelectedRegion) {
+      onChange("auckland");
+    }
+  });
```
andycarrell, line 36: Checks initial value is a valid region, if not, defaults to `auckland` - chosen it to work like this because:
- We default it to auckland in the parent
- The region select defaults to auckland being selected.

This makes it consistent 🙏 - we could add a disabled option, which could be selected by default? That would mean we wouldn't have to run this check 🤔 

image

andycarrell: Actually we can do 1 better, a hidden option will show in the 'select' but not in the options list. And only when the query param is invalid 🤔 I think I'm gonna do that 🙏

app/components/RegionDropdown/RegionDropdown.tsx
```diff
@@ -31,6 +32,7 @@ export const RegionDropdown = ({
         value={selectedRegion}
         onChange={(event) => onChange(event.target.value as RegionOptionId)}
       >
+        <option hidden>Select DHBs</option>
```
andycarrell, line 35: Handles the case where RegionId doesn't match any of the given options

app/components/RegionDropdown/RegionDropdown.tsx
```diff
@@ -16,7 +17,7 @@ const options: Option[] = [
 ];
 
 type RegionDropdownProps = {
-  selectedRegion: RegionOptionId;
+  selectedRegion: RegionOptionId | string | undefined;
```
andycarrell, line 20: Since Region option id is coming from query params it can be any given string (or undefined)

app/hooks/useFreeze.ts
theramis: > Note: these aren't actually used, but could well be useful in a future PR - should I wait and add them when needed?

I don't really know these hooks but I generally go with YAGNI. 
Not fussed, whatever ya'll think is best

andycarrell: Had to push another change anyway so removed lol

jishaal:
Haven't had a good chance to read through properly. Imo merge if/when happy and I'll read properly into it tomorrow :)

andycarrell:
Conflicts now anyway hehe will wait for tomorrow anyway 🙏

app/components/RegionDropdown/RegionDropdown.tsx
```diff
@@ -40,8 +41,6 @@ export const RegionDropdown = ({
         })}
       </select>
       <div className={styles["dropdown-caret"]}>
-        {/* <span className={styles["caret-left"]} />
```
jishaal, line 43: Imo we may as well delete the CSS for this as well

jishaal:
> Update the region dropdown to properly reset the selected value on mount if it's not correct. @jishaal - this handles invalid attributes passed down. Let me know what you think, it's quite defensive but prevents an inconsistent UI

Nice works well! Hopefully unlikely folks will end up in this situation but good to cater for

jishaal (approved):
Noice this looks good to me!

app/components/RegionDropdown/RegionDropdown.tsx
andycarrell: Already gone ;)

PR #166, opened 2021-11-16
RFC - Avoid committing env vars

andycarrell:
### Motivation
Because Next looks at both .env.local & .env.development, we can gitignore one (.env.local) and reference variables from there as "Secrets" this means we can store copies of variables locally and not have to worry about modifying a committed file 💃 

We keep .env.development so that we can set defaults like environment & GA etc

_Note: This will require a slight change to your local env file to work_


.env.development
```diff
@@ -0,0 +1,7 @@
+NEXT_PUBLIC_ENVIRONMENT=local
+NEXT_PUBLIC_GA_ID=G-LGFTVFGLKX
+
+# Get this from this message https://discord.com/channels/658523382442360866/903771050385502239/904216952141152328
+DISCORD_WEBHOOK=$SECRET_DISCORD_WEBHOOK
+GITHUB_PAT=$SECRET_GITHUB_PAT
+API_KEY=$SECRET_API_KEY
```
theramis, line 7: I'm guessing the slight change you mentioned would be to set `SECRET_DISCORD_WEBHOOK`, etc in the `.env.local` file?

Is it worth adding a comment here to say that peeps should do that?

theramis (approved):

theramis (approved):

PR #168, opened 2021-11-18
Track ga event for refresh button

jishaal:


app/components/LatestDataRefresh/LatestDataRefresh.tsx
```diff
@@ -41,7 +42,10 @@ export const LatestDataRefresh = ({
     >
       <Button
         className={cx(styles["refresh-button"], "animate-fade-in")}
-        onClick={() => router.reload()}
+        onClick={() => {
+          trackEvent({ action: "refresh-latest-data" });
```
jishaal, line 46: Should it be `refresh_latest_data` to follow GA's convention? 🤷🏼‍♂️

finnclark (approved):

app/utilities/trackEvent.ts
```diff
@@ -0,0 +1,16 @@
+type GTagEvent = {
+  action: string;
+  category?: string;
+  label?: string;
+  value?: number;
+};
+
+const trackEvent = ({ action, category, label, value }: GTagEvent): void => {
+  window.gtag("event", action, {
```
andycarrell, line 9: _Probably_ want to do a `window` undefined check here just in case ;)

finnclark: Whoopsie sorry I just merged it last night. Want to create a new PR and we can chuck it in before the button goes up later today?

andycarrell: That's all good! It's probably being extra defensive 💃 but yeh will do 1 sec

PR #169, opened 2021-11-18
Add percentage increased for each dhb doses

theramis:
## Changes 
- Added the percentage increase for each dhb
- Removed the prop for doses change from before (I'm assuming it was leftover from before)
- Made a call to not show any percentage increase if it's less than 0.1 (@finnclark you alg with that?)
- Moved the tick to the left of the number as per design

**Expectation**
image

**Reality**
image

## TODO

Can someone please checkout this branch and fix the CSS 😅 
One day I'll learn it I promise.


finnclark (approved):
LOVE IT

PR #172, opened 2021-11-18
Adds percentage-change-negative class to flip that icon

andycarrell:


theramis (approved):

app/components/DosesDescriptionList/DosesDescriptionList.module.scss
```diff
@@ -40,6 +40,12 @@
   }
 }
 
+.percentage-change-negative {
+  & svg {
+    transform: scaleY(-1);
```
finnclark, line 45: @andycarrell question for later: why scaleY(-1) over rotate(180deg)?

andycarrell: @finnclark sorry just got back round to this. Short story - because the icon is symmetrical about the vertical axis, both are equivalent.

Rotating turns the element about the center point, whereas scale (with a negative number) is a "point reflection" which is a mirror image about the center point. 

For example if the icon was this: ↗️ 
then rotating it would give:          ↙️
But flipping it would give: ↘️

PR #184, opened 2021-11-26
Adds 'date' page, with param - so we can render a page for days in the past

andycarrell:
## Motivation
Since we have the data, it's fairly easy to generate pages for days in the past. This will be useful for our users to be able look back at data over time, but also for us to debug day to day changes.

It works by a url param, `/{year}-{month}-{day}` so `https://www.outoflockdown.co/2021-11-20` will show data for the 20th of November. "today" or dates in the future will redirect to the homepage. 

## Challenges 

How do we ensure that the user can easily tell this isn't up to date information? I've added the relevant dates (updated per your suggestion @jishaal), but it may still be a little confusing. There's also no way (at this point) of navigating to these pages, but that could be done in a future PR.

## Outstanding questions

- Naming: I've just called this page 'date' that kind of works but is _very_ generic
- URL path: Should the date param be prefaced? Something like: https://www.outoflockdown.co/date/2021-11-20 ?
- This is a bit more for us to maintain 😅 - happy to drop this if we'd prefer 🙏 

image

### DIFF
--- a/app/propsGenerators/indexPagePropsGenerator.ts
+++ b/app/propsGenerators/indexPagePropsGenerator.ts
@@ -1,9 +1,13 @@
 import { DateTime } from "luxon";
-import { getAllVaccineData } from "../../shared/vaccineDataStore";
+import {
+  getAllVaccineData,
+  getVaccineDataForDate,
+} from "../../shared/vaccineDataStore";
 import { getLatestFetchyRun } from "../../shared/githubDataStore";
 import {
   DhbVaccineDoseData,
   DhbName,
+  NzTimeIso,
 } from "../../shared/types/VaccineDataTypes";
 import { CONSTANTS } from "../constants";
 import {
@@ -12,6 +16,8 @@ import {
   IndexPageProps,
 } from "../types/IndexPageProps";
 import { createCache } from "../../shared/cache";
+
+export type DatePageProps = Omit<IndexPageProps, "lastCheckedAtTimeUtc">;
 
 function calculateDosePercentage(
   dosesCount: number,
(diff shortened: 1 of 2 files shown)


pages/[date].tsx
```diff
@@ -0,0 +1,132 @@
...
+} from "../app/components/Page";
+import { RegionDropdown } from "../app/components/RegionDropdown";
+import { DhbRegionId } from "../app/types/IndexPageProps";
+import {
+  fetchDatePageProps,
+  DatePageProps,
+} from "../app/propsGenerators/indexPagePropsGenerator";
+import { DhbsVaccineDoseDataList } from "../app/components/DhbsVaccineDoseDataList";
+
+const Date: React.FC<DatePageProps> = (props: DatePageProps) => {
```
Comment (andycarrell) [inline pages/[date].tsx:22]: Naming: doesn't really matter but could be diff?

pages/[date].tsx
```diff
@@ -0,0 +1,132 @@
...
+  return DateTime.fromObject(
+    { day, month, year },
+    { zone: "Pacific/Auckland" }
+  ).toISO();
+};
+
+export const getServerSideProps: GetServerSideProps<DatePageProps> = async ({
+  query,
+}) => {
+  const queryNzIso = queryDateToNzIso(queryParamAsString(query.date));
```
Comment (andycarrell) [inline pages/[date].tsx:110]: This should be `params` 🤦 will update

jishaal:
Wonder if we want to show these different dates? i.e should we only show the valid as at date rather than including the date we fetched it?

image

andycarrell:
Yeh @jishaal I think that's probably a good idea. My thinking was maybe the link (to MoH) should be different then? Because they'll click that and be taken to a different page?

app/propsGenerators/indexPagePropsGenerator.ts
```diff
@@ -25,6 +31,33 @@ const getCachedLatestFetchyRun = createCache(getLatestFetchyRun, {
   ttlInSeconds: 5 * 60,
 });
 
+export async function fetchDatePageProps(
```
andycarrell, line 34: TODO: extract these functions to separate files _or_ rename file - will do in a separate PR

theramis:
Sooooooo I'm going to close. 
Sorry boys it's annoying me in the PR tab 😅

jishaal:
Lol, good times 😉

PR #217, opened 2022-02-17
Fixes fetchy by updating regex per site update

andycarrell:
🙏 Should fix it!


jishaal (approved):
Ah legend thanks @andycarrell

PR #218, opened 2022-02-21
Fixes fetchy by updating regex per site update (again)

andycarrell:
Looks like they updated the site again 😅 

I've also regenerated the data for the last few days that it was off for

### DIFF
--- a/shared/vaccineDataStore.ts
+++ b/shared/vaccineDataStore.ts
@@ -31,7 +31,7 @@ export const getAllVaccineData = async (
   const files = await fs.readdir(CONSTANTS.vaccineDataFolder);
   const filesToParse = files.sort().reverse().slice(0, maxItems);
 
-  return await Promise.all(
+  return Promise.all(
     filesToParse.map((f) => {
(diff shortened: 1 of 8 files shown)


shared/vaccineDataStore.ts
```diff
@@ -59,7 +59,7 @@ export const deleteAllVaccineData = async () => {
 export const getAllRawVaccineSites = async () => {
   const files = await fs.readdir(CONSTANTS.rawSiteFolder);
 
-  return await Promise.all(
+  return Promise.all(
```
theramis, line 62: Why did this change?

theramis (approved):

shared/vaccineDataStore.ts
andycarrell: Minor refactor while in the area - since the caller of this function will await the returned promise, we don't need to "double" await it. 

Can remove if you'd prefer?

andycarrell:
Hmmmm just looking at the preview something isn't right 😅

andycarrell:
Superseded!

PR #219, opened 2022-02-26
Hardcode old date until regex is fixed

theramis:


andycarrell:
I think it may even need to be one day before this? The numbers look out of whack (because I think the HTML has changed)

image

PR #220, opened 2022-02-27
Added date based regex and some basic tests

theramis:
## Changes
It's not my best but it'll do

- Added some a basic test to check if all files have the write data (I wanted more thorough stuff but couldn't think of a nice way of doing it and this was enough to tell me the regex was working so kept that)
- The name page doesn't have # to 90 percent so calculating it for all new pages
- regen the data for the new pages 
- Running tests on PR + deploy build. This way if the page changes again and we couldn't parse it "properly" (i.e. if we don't have 21 dhbs) the deploy would fail. We don't have notifications for this but atleast the site won't break 😅


fetchy/extractors/vaccinationDataPerDhbExtrator.ts
```diff
@@ -1,19 +1,65 @@
...
+    dailyUpdatedNzTimeFromHtml.year,
+    dailyUpdatedNzTimeFromHtml.month,
+    dailyUpdatedNzTimeFromHtml.day
+  );
+
+  if (rawHtmlDate <= new Date(2022, 2, 15)) {
+    return extractDhbData_Including_and_Before_15_02_22(rawHtml);
+  }
+
+  if (rawHtmlDate <= new Date(2022, 2, 16)) {
```
andycarrell, line 22: Broadly, I like this approach - I think though that we should copy _the whole file_ - RN we know that the regex and the shape of the table have changed. In the future it could be more, and we'll have to come through and refactor this file again. Conversely, if we copy the whole file per date range, we can 1. avoid the `null` checks 2. tie the regex to the extractor. HMU if you want

netlify.toml
```diff
@@ -1,5 +1,5 @@
 [build]
-  command = "npm run build"
+  command = "npm run test && npm run build"
```
andycarrell, line 2: Technically, atm the tests verify fetchy, which is "deployed" as soon as the PR is merged. I think running tests here is a bit unnecessary (and even prevents us deploying even if it's only fetchy that's broken) I probably _wouldn't_ run them here, and rely on the PR check

.github/workflows/verify-fetchy.yml
```diff
@@ -22,6 +22,8 @@ jobs:
           cache: "npm"
       - name: Install dependencies
         run: npm ci
+      - name: Run Tests
+        run: npm test
```
andycarrell, line 26: Big fan of this! In terms of verification, it will be slightly better coverage if we create a whole new action to run on PR - that will run fetchy & test in parallel, and ensure that we verify fetchy even if tests fail ;)

fetchy/extractors/vaccinationDataPerDhbExtrator.ts
theramis: hahahaha I actually did it this way first. You mean like a strategy pattern right?

I assumed people wouldn't want duplication so I put it all in one place even though I prefer strategy pattern too haha. 

Will do that :)

netlify.toml
andycarrell: Yup let's do it 👍 I think ultimately unit tests are low cost / run fast - there's no reason not to run them then.

fetchy/extractors/vaccinationDataPerDhbExtrator.ts
andycarrell: If that's what it's called ;) haha yeh I think the duplication is fine :D

fetchy/vaccineDataExtractor.test.ts
```diff
@@ -0,0 +1,14 @@
+import extractVaccineData from "./vaccineDataExtractor";
+import { getAllRawVaccineSites } from "../shared/vaccineDataStore";
+
+test(`Extracts vaccine data per dhb from all html files`, async () => {
```
andycarrell, line 4: :nit: Template literals are unnecessary here

```suggestion
test("Extracts vaccine data per dhb from all html files", async () => {
```

fetchy/extractors/vaccinationDataPerDhbExtractors/extractorIncludingAndBefore16_02_22.ts
```diff
@@ -0,0 +1,63 @@
...
+    secondDosesPercentageRaw: match.groups!.secondDosesPercentage,
+    secondDosesTo90Percent: secondDosesTo90Percent,
+    totalPopulation: totalPopulation,
+  };
+};
+
+const calculateDosesToReach90Percent = (
+  currentDoses: number,
+  totalPopulation: number
+): number => Math.max(Math.round(totalPopulation * 0.9) - currentDoses, 0);
```
andycarrell, line 63: :minor: can this can be extracted to the utilities file? and then re-used across extractors that need it?

fetchy/extractors/vaccinationDataPerDhbExtractor.ts
```diff
@@ -0,0 +1,27 @@
+import { DateTime } from "luxon";
+import { DhbVaccineDoseData } from "../../shared/types/VaccineDataTypes";
+import { extractDhbData_Including_and_Before_15_02_22 } from "./vaccinationDataPerDhbExtractors/extractorIncludingAndBefore15_02_22";
```
andycarrell, line 3: :nit: Prefer not to have underscores in this name, we can do:

```suggestion
import { extractDhbDataIncludingAndBefore15Feb22 } from "./vaccinationDataPerDhbExtractors/extractDhbDataIncludingAndBefore15Feb22";
```

fetchy/extractors/vaccinationDataPerDhbExtractor.ts
```diff
@@ -0,0 +1,27 @@
...
+import { extractDhbData_Including_and_Before_16_02_22 } from "./vaccinationDataPerDhbExtractors/extractorIncludingAndBefore16_02_22";
+import { extractDhbData_Including_and_After_17_02_22 } from "./vaccinationDataPerDhbExtractors/extractorIncludingAndAfter17_02_22";
```
andycarrell, line 5: :minor: Since we'll have to rename this if / when the regex changes again, I think we should call it something a bit more generic like 'current'. Then cycle through renaming it if needed.

```suggestion
import { extractDhbDataCurrent } from "./vaccinationDataPerDhbExtractors/extractorCurrent";
```

andycarrell (approved):

